PR #2, opened 2015-03-04
Janitorial fixes for GitHub

danielhjames:
Just adding a gitattributes file to help CRLF issues and a pretty README.


finwe:
Could you squash last two commits?

danielhjames:
No problem, that squash is done. I agree it is tidier now.

PR #8, opened 2015-04-09
Remove redundant PHP endings.

ram4nd:


ram4nd:
These three are hopefully merged, the font issue should be a separate issue. It includes updating of font documentation, etc.
- Remove redundant PHP endings.
- Small bugfix, preg_replace return was not caught.
- Performance: implement static cache for mpdf function ConvertColor.

PR #12, opened 2015-04-28
Chmod 777 dirs ttfontdata, tmp, graph_cache after install

SilverFire:
After installing MPDF with composer, it does not work because of insufficient permissions on temp dirs. People should read Readme to make it work, but when MPDF is a dependency of another package, this may cause confusions for the developer.
Let's make default behavior to make MPDF work just after `composer install`.

If someone want to set another permissions - it will be OK, because after `composer update` custom user's permissions will stay unchanged.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,5 +22,12 @@
 	},
 	"autoload": {
 		"classmap": ["mpdf.php", "classes"]
-	}
+	},
+	"scripts": {
+	        "post-install-cmd": [
+	            "php -r \"chmod('./ttfontdata', 0777);\"",
+	            "php -r \"chmod('./tmp', 0777);\"",
+	            "php -r \"chmod('./graph_cache', 0777);\""
+	        ]
+    	}
 }


finwe:
Nice catch, thanks, but 777 is maybe too open. What about changing it to 775?

Could you also shorten the commit message? I see no reason it should contain whole description of the issue.

SilverFire:
The popular way of web server configuration is to run web-server from user `www-data`, `www`, etc., and to store files from another user:group, for example `silverfire:silverfire` - in this case 0775 is not enough. 

About the commit message - that was accidentally, sorry. Lets finish our discuss about the proper chmod and I'll kill this PR, commit and re-create all.

finwe:
OK, fine, lets keep `777`. Re-creating should not be necessary, just amend the commit and force push.

SilverFire:
Great tip, thanks. Done.

finwe:
I've unified whitespace to tabs, removed trailing whitespace, amended and merged.

PR #14, opened 2015-05-18
Updated classes of FPDI

JanSlabon:
Updated classes of FPDI to their current version and updated the code in mPDF respectively.
Additionally missing example pdfs were added.


JanSlabon:
Encountered problems with foreign PDF documents... need to withdraw this pull request.

PR #16, opened 2015-05-18
Updated classes of FPDI

JanSlabon:
Updated classes of FPDI to their current versions and updated the code
in mPDF respectively. Additionally missing example pdfs were added.


blueliquiddesigns:
I use the mPDF FPDI implementation daily and this looks great. 

A couple of questions:

1. Is there a changelog for FPDI for review?
1. What's the minimum PHP versions requirements for the new version? mPDF still supposedly supports PHP4 (which in future I think should be increased to at least 5.3, but probably higher if it is to be modernised)

Finally, any chance the SetAsign team would ever consider open sourcing their commercial FPDI PDF-Parser (http://www.setasign.com/products/fpdi-pdf-parser/details/). The 1.4 PDF import restrictions for FPDI can be a real buzz kill sometimes.

JanSlabon:
To your questions:

to 1) The whole changelog is available [here](https://www.setasign.com/products/fpdi/downloads/). Anyhow I didn't implemented all changes in mPDF because it is very geared by some changes Ian did. This update simply brings the used classes to their latest version. Actually it is also possible to simply update the files if a new FPDI release is done (at least in the 1.5 branch).

to 2) Actually FPDI is PHP 5.2 (latest) compatible.

The separate parser will not go open source (in view to its license). License owners, as you Jake, have access to the source code :-)

Anyhow you should keep in mind that the licenses of mPDF is not compatible to ANY proprietary license.

blueliquiddesigns:
Thanks for the feedback, Jan. I'll check out the release notes when I go to merge in these updates and test them with the mPDF version I have. 

I thought it might have been PHP5.x+. If that's the case before these are merged in the PHP4.x compatibility currently in mPDF should be removed and the minimum PHP version requirement bumped up. 

If SetAsign owes the copyright to the 1.5+ PDF parser they can always change, or dual license, the code ;) It wouldn't be the first time a commerce software has become open sourced, and then the licenses would be compatible. I'm just throwing the idea out there. It would be great if it ever was open sourced but understandably that could effect your commercial interests - although it hasn't stopped other paid software becoming open sourced and still flourishing (http://wptavern.com/gravityview-is-now-public-on-github).

JanSlabon:
Jack, I'm sorry but GPL is not an option for us, sorry.

blueliquiddesigns:
It's always worth asking I say! I appreciate the quality of work the Setasign team does in bring PDF manipulation to PHP. Keep up the good work.

JanSlabon:
I close this pull request in favor of #47.

PR #17, opened 2015-05-19
Throw Exceptions\mPDFException instead of die() in mpdf.php Class

Gudabu:


TomK:
This is really important for any production environment. Now that errors are handled in one place (by mPDF::Error - #52), that function should be updated to throw an exception instead of die

finwe:
Thanks for the pull request, I've built it a bit differently by myself - the `Error` method is now removed and the `MpdfException` is thrown directly on the line of error.

I have also changed `die` for `throw` in additional classes and scripts.

PR #24, opened 2015-06-09
Fix LEDGER and TABLOID

Verron:
Currently, the LEDGER and TABLOID sizes are specified in millimeters while others are specified in points. This PR changes the sizes for LEDGER and TABLOID to be specified in points as well.

Reference:
http://www.gnu.org/software/gv/manual/html_node/Paper-Keywords-and-paper-size-in-points.html


finwe:
Merged, thanks!

PR #30, opened 2015-06-14
Set up and Configure PHPUnit and write baseline tests for the mPDI (FPDI) classes

blueliquiddesigns:
Unit Testing Class in PSR-2 Standard


finwe:
Thanks for the first tests ever.

A brief review:

* Lets stick to naming test *Test.php, therefore test-MPDI.php would be MPDITest.php. 
* Name of the class should equal to name of the file, therefore ```MPDITest```
* phpunit.xml file should be in the tests directory (adjust bootstrap accordingly)
* convertErrorsToExceptions and other xml settings are true by default, therefore unnecessary
* use composer for autoloading in bootstrap.php
* add phpunit to require-dev in composer.json

finwe:
Plus please rebase the PR on master.

blueliquiddesigns:
Thanks for the feedback. 

I've gone through and believe I have fixed up most of the issues but need some additional feedback regarding others. 

> I am unable to run the only test on its own - there are multiple unrecognized constants

You would have needed to execute `phpunit` from the mPDF root directory, rather than the `/tests/` directory. However after your feedback you now need to run `../vendor/bin/phpunit` from the `/tests/` directory (after running `composer install` - the details are in the updated README.md file for clarity).

> convertErrorsToExceptions and other xml settings are true by default, therefore unnecessary

I found the `colors` option wasn't on by default so it was left in.

> keep sublime files and folders in your local .gitignore file

The point of having a `.gitignore` file is to keep unnessisary files from being committed. Excluding a number of common editor-specific temporary files will make it simpler for developers to clone the repo and create PRs in the future, plus it will be easier for the person doing the review process.

> remove trailing whitespace in separate commit

I'm not sure which commit you are talking about. My editor is set to remove all trailing whitespace by default.

> name directory with example data data - as there can be multiple types of content

I've also included a subdirectory (pdfs) to help separate the test data in future.

> Why do you export-ignore the .gitignore file?

It's my understanding that this option is used when running `git archive`. Since creating an archive removes the `.git` directory and is no longer under version control the `.gitignore` file isn't needed. This could be further expanded to the `.gitattributes` file. 

> Plus please rebase the PR on master.

Could you explain why this is necessary? The PR is for the `mpdf/mpdf` master branch but I didn't think it mattered if I created a new branch off the master in my local repo and used that for the PR.

finwe:
Thanks for a swift correction!

> However after your feedback you now need to run ../vendor/bin/phpunit from the /tests/ directory

This now works perfectly.

> I found the `colors` option wasn't on by default so it was left in.

Sure, excellent. I believe the `<testsuites>` section is also really not necessary.

> remove trailing whitespace in separate commit

Yes, removing trailing whitespace is a good thing to do, but by saving a file to which you won't do any notable changes, it messes up the diff so it is much harder to point out what and where the particular change was. It would be much better to remove the whitespace in separate commit and make the change itself in following commit.

But whatever, it is only README.md in this case and it doesn't matter as much.

> keep sublime files and folders in your local .gitignore file

The point of repository .gitignore file is to prevent comitting files the application itself uses, creates and/or changes. It is not possible to keep all "common" IDE-specific files in the repository gitignore, it should, IMHO, be the responsibility of the developer and his/hers local .gitignore file, to keep own IDE files out of any repository.

Also, it is much better for clarity of the .gitignore file - listed files point to paths which the project itself uses but does not rely on them to be in the repository.

> Plus please rebase the PR on master.

I'd like to manage this repo according to [a simple git branching model](https://gist.github.com/jbenet/ee6c9ac48068889b0912) - this has many advantages: no merge hell, simple and plain history, clear diffs between adjacent commits. This requires all branches for PR to be built (or rebased onto) the topmost master commit. I can do this myself for this particular PR as it does not change any existing files except for readme, but for future PRs I'd like the participants to do that.

And a few additional ones:

* composer.lock is a perfect example of a file which should not be committed and should be in the project .gitignore file.
* please remove the <testsuites> portion of phpunit.xml file as hinted above.
* I think that all commits of the branch could (and should) be squashed to one or at most two commits - once again for clarity and pure history.

blueliquiddesigns:
Thanks again for the quality feedback.

>Sure, excellent. I believe the `<testsuites>` section is also really not necessary.

When I removed this the tests wouldn't run so I've left it in.

> composer.lock is a perfect example of a file which should not be committed and should be in the project .gitignore file.

I've removed the IDE-specific files and included composer.lock in the .gitignore

> I'd like to manage this repo according to a simple git branching model

This looks like a good way to do things. I've rebased my branch into a single commit. For future reference, if I hadn't already done a PR, would I then pull down the master from the orgin so it was up to date and rebase the branch into the master (after which I would then do a PR from the master branch)?

finwe:
Thanks again, merged.

blueliquiddesigns:
Great. 

If you could just let me know the following I can ensure I correctly prepare the PR for merging in future:

> For future reference, if I hadn't already done a PR, would I then pull down the master from the orgin so it was up to date and rebase the branch into the master (after which I would then do a PR from the master branch)? Or can I still do this process [meaning, can I open a pull request from a new branch for discussion, get it correct, rebase into a single commit on the new PR branch and then rebase into the master branch for merging]?

finwe:
Here is a nice how-to to keep your fork up to date with upstream https://robots.thoughtbot.com/keeping-a-github-fork-updated

You would do this procedure before starting work on a new feature/fix, and once again just before submitting pull request. If your work takes some longer time and upstream changes in the meantime, is is recommended to rebase every once in a while to keep potential conflicts as easy to resolve as possible.

Is this a sufficient description of this workflow?

blueliquiddesigns:
Thanks Matěj for your description and patience in this matter. Yes. I believe I understand the workflow now.

Keep up the good work managing the mPDF repo.

PR #31, opened 2015-06-15
Base formatting of base classes

finwe:


TomK:
Hi, I was just rebasing #35 and updating it to maintain php4 compatibility, but its hard to maintain the existing code style of some of these classes.  So i figured i could lend a hand with styling first and redo my PR once that is complete.
I just wondered what standards you are aiming for (PSR?), if perhaps I can configure PhpStorm to match your preferences and run auto-formatting for speed.
Thanks, Tom

PR #35, opened 2015-06-15
PHP 4 constructors are deprecated in PHP 7

TomK:
I'm not sure what your target php version is, but this PR should address the break in compatibility with PHP 7.
If you want to retain PHP 4 compatibility you can probably do this by leaving the original methods, and calling them from new __construct methods.  I have not done this as PHP 4 was [end of life](http://php.net/eol.php) on 7 Aug 2008.

### DIFF
--- a/classes/barcode.php
+++ b/classes/barcode.php
@@ -50,6 +50,10 @@ class PDFBarcode {
 	protected $daft;
 
 	public function __construct() {
+		$this->PDFBarcode();
+	}
+
+	public function PDFBarcode(){
 
 	}
 
--- a/classes/bmp.php
+++ b/classes/bmp.php
@@ -4,10 +4,13 @@ class bmp {
 
 var $mpdf = null;
 
+function __construct(&$mpdf) {
+	$this->bmp($mpdf);
+}
+
 function bmp(&$mpdf) {
 	$this->mpdf = $mpdf;
 }
-
 
 function _getBMPimage($data, $file) {
 	$info = array();
--- a/classes/cssmgr.php
+++ b/classes/cssmgr.php
@@ -9,6 +9,9 @@ var $cascadeCSS;
 var $CSS;
 var $tbCSSlvl;
 
+function __construct(&$mpdf) {
+	$this->cssmgr($mpdf);
+}
 
 function cssmgr(&$mpdf) {
 	$this->mpdf = $mpdf;
--- a/classes/directw.php
+++ b/classes/directw.php
@@ -4,6 +4,10 @@ class directw {
 
 var $mpdf = null;
 
+function __construct(&$mpdf) {
+	$this->directw($mpdf);
+}
+
 function directw(&$mpdf) {
 	$this->mpdf = $mpdf;
 }
--- a/classes/gif.php
+++ b/classes/gif.php
@@ -26,6 +26,11 @@ class CGIFLZW
 	///////////////////////////////////////////////////////////////////////////
 
 	// CONSTRUCTOR
+	function __construct()
+	{
+		$this->CGIFLZW();
+	}
+
 	function CGIFLZW()
 	{
 		$this->MAX_LZW_BITS = 12;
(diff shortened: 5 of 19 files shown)


finwe:
Thanks. 

This change is a way ahead of current state of things. Firstly, I would like to have https://github.com/mpdf/mpdf/pull/31 merged before changes like this. I believe mPDF officialy supports PHP 5.2, however it can be run on PHP 4 when not using many advanced features. In the first round of code cleanup, I would like to retain this compatibility.

This PR could also remove explicit reference operator `&` from constructor mpdf parameters, and add class type hints.

Heads up: see classes/otl.php - you have supposedly converted the source to UTF8, resulting in loss of some characters.

TomK:
Thanks @finwe, I noticed the bad encoding too after I submitted the PR.  I have manually reinstated these characters.

I did see a few commits on your reformatting branch and wondered if i should have forked that instead.  How would you like me to proceed with this?  I'm happy to dereference the constructor arguments within this PR, although a merge after #31 would effectively be rewriting it anyway.  I can hold off on this for now if you'd prefer?

finwe:
Yes please, hold on, I will post here when #31 is merged, then you could rebase or start over if simpler.

Thanks!

TomK:
I've just done a little testing, it seems to be a fatal error to maintain php4 style constructors and call them from php5 __construct methods.. You cannot have both style constructors. 

So i think we would have to deprecate php4 in order to support php7.

blueliquiddesigns:
I agree with Tom about removing PHP4 support. Based on his feedback, it sounds like if you wanted to support both PHP7 and PHP4 you would have to have both constructs call a third method (instead of calling the PHP4 construct from the PHP5 construct). Then there is testing to consider. Anyone have PHP4 compiled? We'd need to run tests in multiple versions...

Since PHP4 is about to be 7 years old and PHP7's 2nd beta was released this week I'd recommend we look to moving forward on this. 

With this change it also would be good to decide on the minimum supported PHP version mPDF should support in the future. I'd recommend PHP5.3.2 which will allow us to take advantage of namespaces, autoloading, closures and SPL. Depending on where you look, it is also still one of the most-used versions of PHP (http://w3techs.com/technologies/details/pl-php/5/all) - although WordPress (which boasts running 25-ish% of the web) stats show PHP5.4 is the most popular (40% vs 33%) https://wordpress.org/about/stats/.

TomK:
@blueliquiddesigns I just had a quick play, it seems if you declare the new constructor before the old constructor, then there is no strict standards error.  So theoretically we could run both constructors after all.
Ideally I'd like to get this patched in and versioned before php7 is released.  Although i'm aware we may be waiting for the code formatting to be completed, I do feel this should be prioritised over formatting.

This works fine:
```
class testClass {
  public function __construct() {}
  public function testClass() {}
}
```

This produces a strict standards warning:
```
class testClass {
  public function testClass() {}
  public function __construct() {}
}
```

TomK:
i've updated the PR with a squashed commit. and introduced some additional constructors for classes which did not have php4 constructors anyway.  Should be more consistent now.

Some cleanup occurred on one of the files, whitespace trimming and some character encoding.  Not sure how to easily remove that, any thoughts?

TomK:
Ok, i've tidied it up. I just checked out the upstream files and recreated them in nano instead of PHPStorm.  I've tried to maintain the individual files' coding standards where possible.

danielhjames:
Thanks for this work, Tom! I am very interested in seeing the speed-ups we might get from PHP 7 when producing long books with lots of complex font embedding, e.g. Arabic and CJK. Currently a 500 page book in Arabic can take up to ten minutes to render when using font subsetting.

TomK:
@danielhjames you're very welcome.  Our initial PHP7 trials across the board have been really promising.  It seems to bring it in line with HHVM.
I'm hoping the ownership issues of mpdf can be solved soon, there seems to be a lot of good developers itching to improve it and implement modern standards :)

danielhjames:
In terms of ownership, Ian will always retain copyright in his work, but happily the GPL is designed for exactly this situation. If mPDF had been released as freeware under a license which did not permit source redistribution, the community would be really screwed!

TomK:
@finwe Let me know if this needs rebasing again.  Also, contrary to the title, this PR should maintain backwards compatibility as i have not removed the php4 constructors.  However i have not tested it on PHP4.

finwe:
Hi, could you please recreate this on top of development branch? It already has code style issues integrated into.

PR #43, opened 2015-08-11
Fix for list_item_marker undefined variable

alexyoungs:


finwe:
Fixed as part of 348d313, but thanks!

Issue #15, opened 2015-05-18
License issue: GPL v2 vs. ASL 2.0


JanSlabon:
It's not really clear under which GPL version mPDF is released. On [this page](http://www.mpdf1.com/mpdf/index.php?page=Licence) it seems like it is release under both. In the LICENSE.txt version 2 is declared. Somewhere in the source code only "GPL" without a version is used. So I guess it's version 2 atm.

Anyhow, mPDF includes code which is released under the "Apache Software License 2.0" (classes from FPDI) which is [incompatible to GPL v2](https://www.gnu.org/licenses/license-list.en.html#apache2) while it is [compatible to GPL v3](http://www.apache.org/licenses/GPL-compatibility.html).

This issue exists for several years now, but as I just updated the classes from FPDI (#16) I though it's time to move at least to GPL v3?

danielhjames:
Hi Jan, having reviewed the licenses I think this is a genuine issue. Only Ian, the copyright holder of mPDF, could make that license change.

JanSlabon:
Hi Daniel,
Ok, I just saw that you're active on the mPDF forum, too. Could you open a thread there which points to this issue? Time ago I tried to contact Ian by email without success...
Thanks!
Jan

danielhjames:
Hi Jan, if I get to speak to Ian I will mention licensing, as in my opinion the GPL v2/Apache mix makes the code difficult to redistribute, and I doubt that was the intention.

blueliquiddesigns:
I put a post over in the mPDF forums asking Ian to chime in about the licensing issue (http://www.mpdf1.com/forum/discussion/2451/gpl-licensing-clarification). I also asked about open sourcing the documentation.

skreutzer:
There's yet another problem, the “SaaS loophole”: the GNU GPL licenses are insufficient for code which runs on servers (see http://radar.oreilly.com/2007/07/the-gpl-and-software-as-a-serv.html), therefore the GNU AGPL 3 (+ or, at your option, any later version) would be much better. It might be that the ASL2 is incompatible with AGPL3 or later, but if existing mpdf code would be licensed under GPL3, new code modules could be incorporated under the AGPL3 (https://en.wikipedia.org/wiki/Affero_General_Public_License#Compatibility_with_the_GPL. Would an “or later” clause be possible, too?), effectively making a combined package seem to be AGPLed, as the GPL3 conditions would be fulfilled automatically where the AGPL3 gets fulfilled with its additional conditions.

[...]

PR #47, opened 2015-10-13
Update FPDI to 1.6.0

JanSlabon:
Removed FPDI classes and add FPDI via a require entry in the composer.json file.
FPDI 1.6.0 is released under MIT now which is compatible with the GPL 2.0 of mPDF.


blueliquiddesigns:
I think the autoloader should be included at the top of the mPDF.php file. Users only need to `include("mpdf.php");` and not worry about the vendor autoload.php file.

blueliquiddesigns:
Disregard my last comment (now removed). I applied this patch and tested it on a couple of different PDFs without issue. Nice work, Jan.

JanSlabon:
Before this change it was possible to run mPDF without composer. Now you need to have it installed with composer if you want to use the PDI features. So it may be a BC break but this way the code was reduced by some files and an update of FPDI shouldn't be that hard :-)

blueliquiddesigns:
This approach works out great [for my particular project](https://github.com/GravityPDF/gravity-forms-pdf-extended/tree/beta) as we're already using composer to install mPDF. Although I'm not sure if making a composer dependency on the main package and breaking BC is the right move, but ultimately @finwe will need to decide which way to go.

JanSlabon:
Don't you mean @danielhjames ?

blueliquiddesigns:
The last pull request I did was merged by @finwe and looking at this repos commit history he's handled most of the commits.

JanSlabon:
Ahh ok, I didn't noticed that :-)

finwe:
Hi, gentlemen,

Thanks for all your work. I'd be very happy to merge this, however I tend to stick as close to the official release as possible until Ian Back gives us his final and official blessing to move github version of mPDF forward (which would mean us to be able to issue our own new versions reflected on the official website).

I haven't seen Ian for a while on the official forum and can't find his official contact address so I could reach him with this request. I am reluctant to do so on the forum, but I guess I will have no other option.

blueliquiddesigns:
A change will have to be made regardless of Ian's blessing due to incompatible licenses with earlier versions of FPDI (see #15). I've also read Ian's comments on the forum about his feelings on using GitHub. With his [lack of input](http://www.mpdf1.com/forum/discussion/2451/gpl-licensing-clarification) about resolving the licensing problem it gives the impression he doesn't want to actively develop mPDF further. We might find we'll make more progress continuing mPDF development ourselves, or forking the project entirely...

finwe:
Yeah, I see the forking as a last resort. I would very much like to continue the original project.

danielhjames:
Hi all, the last time I spoke to Ian, which was some months ago, he indicated that he was in favour of community maintenance of mPDF for versions after 6.0. This is consistent with what Ian has posted in the mPDF forums. What Ian didn't want to do is be responsible for managing a developer community, or have to fix the mess that random code changes might make, which is easy to understand.

What I'd like to propose is that we establish a community development process, with only one person (finwe?) having merge powers into the https://github.com/mpdf/mpdf project in order to maintain consistency. I suggest there would be one branch for fixes to 6.0 which could potentially be the basis of an official 6.1 release, with Ian's approval, as these changes should be relatively easy to review. There would be another branch for a future mPDF 7.0 release which was refactored, with all the improvements we would like to see.

I'm sure that ours is not the only organisation with a strong interest in the future of mPDF. We are actively looking for a developer we can pay to implement new features and fix bugs, for example fixing the widows and orphans issue. We would want these improvements to be part of the community process.

blueliquiddesigns:
Sounds like a great idea.

marclaporte:
@danielhjames: Great!

To facilitate community coordination, I added: https://github.com/mpdf/mpdf/issues/48

Best regards,

M ;-)

finwe:
@JanSlabon Could you please rebuild this PR on top of `development` branch? I see there are some unit tests regarding FPDI, so make sure they are adjusted accordingly.

Also, do we aim with this to the next minor version (6.1) or to next major (7.0)? It might not be a BC break _per se_, when a ZIP distribution package is accordingly produced.

I also came to my mind: what if the fpdi was only a suggested dependency and it was up to the end user to include it in his/hers own `composer.json` file? Is this a good idea or a bad idea?

danielhjames:
I would suggest this update for mPDF 6.1 because of the license conflict, which should be solved as soon as possible. I don't see anything here that changes Ian's design.

As for making FPDI optional, personally I would be fine with that, as not everyone imports PDFs as templates. For those users, it would be less code to deploy in production.

JanSlabon:
I must admit that my time is limited for this issue atm. Anybody who can rebuild this? @blueliquiddesigns? If I remember correct Jake already implemented it in an own repo and didn't he added the tests, too? :-)

blueliquiddesigns:
Leave it with me. I'll work on it this afternoon.

finwe:
Closing (solved with #63)

PR #52, opened 2015-11-01
Change all die to $this->Error

Jako:
If you inherit a new class from mPDF, all those error messages could be handled different by overriding just one method.


finwe:
Thanks, I recreated the PR on top of development branch (merge would be too complicated because of code style fixes)

Issue #36, opened 2015-06-15
Remove PHP4 style constructors


finwe:
This PR does most of the work https://github.com/mpdf/mpdf/pull/35

PR #54, opened 2015-11-03
introduce new constructor style for PHP7 support. Drop php4 support

TomK:
Rebased onto development as requested in #35.  Deprecates #35.


TomK:
Should have mentioned:   Closes #36

finwe:
I have just compiled PHP 4 and tried to run some examples, and I found out that there is a class constant used which causes parse error.

<code><b>Parse error</b>:  syntax error, unexpected ')', expecting '(' in <b>/home/finwe/www/_github/mpdf-mpdf/mpdf.php</b> on line <b>30496</b></code>

So I believe that it is safe to remove PHP4 constructors and "support" entirely. I'd then also vote to add public visibility to new __construct methods.

Could you please edit the PR accordingly?

TomK:
@finwe I have pushed this for mpdf classes now.  I have omitted the fpdi classes as they are covered by @JanSlabon #47.

finwe:
Awesome. I have a few other minor suggestions: could you please:
- remove `&` reference operators from constructor parameters,
- add `mPDF` type hints to `$mpdf` parameters,
- and remove those ugly // CONSTRUCTOR comments?

Then just push the new commits, PR should then reflect them.

TomK:
done, only tests that fail in php7 are related to fpdi classes now.  tests pass in php54 and php56.

finwe:
Merged, thanks!

PR #59, opened 2015-11-05
convert all txt, php, css, and htm files to utf8

### LINKED ISSUE #34: Convert all files to UTF-8
Converting is not that simple as some special characters would be owerwritten - a check and preferably a unit test would be in order. Changed characters can be seen in pull request https://github.com/mpdf/mpdf/pull/21

TomK:
Also adds newlines to the end of some files.  Mainly beacuse it was easier to end with newlines than not to, but also because http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline. 
That being said, all php files should probably have their closing php tag removed if at the end of the line, which would remove the potential "include" output nightmare.
Any Thoughts?

Closes #34.


finwe:
Awesome! Could you please also remove closing ?> tags then in a separate commit? I've done that for the main class and subclasses, but missed collations and examples.

I'll test out the UTF8 version probably in the morning.

TomK:
`sed` is really unhappy about the character encoding on the development branch.  So if it's ok with you, i'll create a PR for removing the closing tags after this one is merged in

finwe:
Merged, thanks!

PR #67, opened 2015-11-19
Development add shrink

fglueck:
<table shrink="1.5"....
shrink the table like autosize but forced


fglueck:
please drop this pull request, the gitclient made some garbage

PR #74, opened 2015-12-07
PHP7: fix SVG handling Fatal error: 'break' not in the 'loop' or 'switch' context

danielhjames:
This small deletion fixes an error seen when rendering SVG as part of a PDF when using PHP 7.1.0-dev (cli) (built: Dec  7 2015 12:44:25)

### DIFF
--- a/classes/svg.php
+++ b/classes/svg.php
@@ -2997,7 +2997,6 @@ class SVG
 				}
 				else if (strtolower($name) == 'stop') {
 					if (!$last_gradid)
-						break;
 					$color = '#000000';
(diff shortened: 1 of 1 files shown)


TomK:
since `$last_gradid` is used elsewhere in that `if` block, it might be better to change it to `if ($last_gradid)` and wrap the whole following code block (lines 3000 to 3038), rather than just setting `$color`.

danielhjames:
Hi @TomK that may well be a better solution, but for now, dropping the break enables mPDF to render this particular document.

PR #77, opened 2015-12-11
Allow integer, float, boolean and certain classes to be passed to WriteHTML()

blueliquiddesigns:
Before this change the PDF wouldn't output anything if a 0 was passed to it.  This issue was prevalent when using the WriteFixedPosHTML() function.

Also allow any numeric value to be passed directly. It will be automatically converted to a string.


finwe:
Cool, I'm OK with this. I've merged this into development with minor stylistic changes.

blueliquiddesigns:
Sweet.

PR #79, opened 2015-12-12
Begin fleshing out the $mpdf->WriteHTML() method 

jakejackson1:
@finwe and others have been doing a great job taking the baby steps needed to modernise this project – from code formatting, to TravisCI integration and better exception handling. Continuing with that trend, this PR is a glimpse at the 'beginning to run (and falling over)' faze of modernising the project.

The idea / goal here is to work through the current mPDF class and refractor it down into simple, easy to follow, unit testable classes and methods. Because of the size and complexity of the current mPDF class this really needs to be done piece by piece.

What you see here is a very early start at optimising and testing the $mpdf->WriteHTML() method. I've purposefully moved the code into its own class so it's easier to build out and manage, but it can just as easily be migrated back into the current mPDF class (or elsewhere) when it's finally completed. 

This PR has only semi-optimised about 1/5 of the WriteHTML() method and it's already taken half a day! It's clear that refractoring the entire class will need to be a team effort. I welcome anyone else to assist with this patch as I won't get to spend much time on it over the holiday period. I've gone through and made comments about bits of code I'm unsure about / could be further optimised. If you have time it would be great to have others review it.


TomK:
Good work!  I wonder if this might be a good way to introduce namespacing and PSR-4 by simply refactoring this class into `\Mpdf\Mpdf`. 
Even better, splitting this into something like `\Mpdf\DocumentWriters\HtmlWriter`. We could have a `WriterInterface` which this and other writers (cell, text, barcode) could all implement. Which means people could easily write extensions.

Splitting the monolithic Mpdf class into manageable namespaced subclasses would make maintenance immeasurably easier.

blueliquiddesigns:
Thanks. 

I like the idea of a writer namespace. I was thinking about an interface but haven't explored the other writing methods in enough detail to determine what should be included. 

We should also consider a more managable file structure as well.

finwe:
I'm closing this as it is outdated. Feel free to propose another PR with up-to-date code.

PR #86, opened 2016-01-06
Php7 support.

ROMOPAT:
Test: https://3v4l.org/KSYtH


finwe:
Merged with slightly more verbose commit message, thanks!

PR #100, opened 2016-03-10
Throw the proper exception error when using FPDI

blueliquiddesigns:
During testing I noticed that `$this->parsers[$fn]->errormsg` always returned blank. This change ensures the current exception message gets returned instead.


finwe:
Does FPDI throw the \Exception class directly, or some child exception class? If the latter, maybe it would be cleaner to catch it specifically...?

blueliquiddesigns:
That particular class looks to throw:

* a normal "Exception"
* a "InvalidArgumentException" exception

finwe:
Merged, thanks!

Issue #32, opened 2015-06-15
Remove all side effects from mpdf.php


finwe:
Separate all direct output from main mpdf class to conform with [PSR-1](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md).

* Separate debug/stats ouptut to its own class (with injectable writer class)
* Separate progressbar to its own class
* Separate warnings (PDFA etc) or just throw an exception with link to error description online

finwe:
See also https://github.com/mpdf/mpdf/issues/27

PR #113, opened 2016-04-06
PSR Logger

bunya:
Hello all!
I want to offer PSR3 logger instead of $debug property. (#32) With injectable writers and filters
Unfortunly logger calls placed in some places only. But I hope it might be make on next step


finwe:
This is nice and will be done eventually, in the 7.x branch after 6.1 release. 

I however have a few problems with this implementation, after only a quick review:

* Custom `Mpdf\Logger` classes are completely unnecessary (Are they just a copy&paste of an existing Psr\Log implementation?) - we can use a dead simple own implementation and let users use eg Monolog for more fine-grained logging.
* in StartProgressBarOutput you introduce a BC break of not throwing an exception and just failing silently when using default NullLogger. This method will be removed in 7.x, so that is not really a serious issue.
* I don't really understand by what key have you added logging calls, they seem a bit random, definitely not covering all cases. Furthermore, I believe we could use two distinct loggers - development logger for development and "user" logger for basic user logging.

I am keeping this open, but the implementation will need to be changed.

bunya:
Good day. Thanks for answer
Well

> Custom Mpdf\Logger classes are completely unnecessary (Are they just a copy&paste of an existing Psr\Log implementation?) - we can use a dead simple own implementation and let users use eg Monolog for more fine-grained logging.

You are right. Many modern projects use internal framework logger or any logger module. If project logger implements PSR3 interface, it may be used in mpdf. But if don't, it may be used `Mpdf\Logger` instead of `print_r` or `var_dump`. Of course, current logger implementation must be extended, but would like to know what kind of writers and filers must be added.

> in StartProgressBarOutput you introduce a BC break of not throwing an exception and just failing silently when using default NullLogger. This method will be removed in 7.x, so that is not really a serious issue.

IMHO logging and exception throwing is parallel things. User may use "silent" catching and write to log or off log (using  NullLogger) and output errors in catch block. But it's really my fail here: logger must call before exception throwing

> I don't really understand by what key have you added logging calls, they seem a bit random, definitely not covering all cases. Furthermore, I believe we could use two distinct loggers - development logger for development and "user" logger for basic user logging.

Sorry. Unfortunly, I rework not all of MPDF code (it's some difficult for understanding for me), but I hope it could reworking step by step. IMHO better make it during separating code into single responsible classes
It's not nesseary to use more then one logger. Separating by log level would better. And user can collect some log level messages only, if would use log filters. For example
* **info** level is for watching document creating stages
* **debug** level is for deep debuging
* **warning** and **error** levels is for watching generating errors in production server

> I am keeping this open, but the implementation will need to be changed.

If I get you rightm I must continue to set logger calls in all of MPDF code. But I hope for your help in it.

And some question. I have a few ideas about MPDF code reworking. If you not opposite to continue collaboration I migth make a few issues and example pull requests

blueliquiddesigns:
> And some question. I have a few ideas about MPDF code reworking. If you not opposite to continue collaboration I migth make a few issues and example pull requests

[This PR](https://github.com/mpdf/mpdf/pull/79) touches on rewriting one of the core mPDF methods. It might not be a bad idea to continue the process and write to include proper logging ect.

finwe:
I don't believe that blind adding of logging calls throughout the mPDF codebase is a good idea.

Let us prepare some kind of RFC for logging in 7.x branch and let the authors of individual changes make their own logging calls when they are needed.

bunya:
@finwe 
Well, it was my first idea. At first add `LoggerInterface` and its simple implementation. After that set logger calls in new code blocks during rewriting (or refactoring)

Now situation
* **info** logger calls in `Mpdf` class is normal. Its tell about generation process
* single **warning** call in `StartProgressBarOutput `. OK, it's not needed, because it's in deprecated block. Must be deleted
* **info** logger calls in `Tag` class is bad idea. IMHO it's **debug** level. Probably to delete or to change.

Generally idea for log levels:
* **emergency** not used
* **alert** not used
* **critical** used if unavailable some module, package etc. May be - in case of PHP system error
* **error** - MPDF error. Used together with throwing `MpdfException`
* **warning** used if minor error without code execution breaking. Used instead of throwing `MpdfException`
* **info** so as now. Information about generation process
* **debug** - debug info: parsing stages result, every element creating (I hope that dividing class `Tag` into separate 'elements' classes is good idea) and etc

And so question: am I continue to set of logger call or clean code and get another task?

bunya:
Ok. Done.
Warning and error calls set, info calls kept stay. Debug calls, IMHO, should set during code refactoring.
Is Unit tests need now or it may be later?

finwe:
Hi, @bunya, thanks for the work you've done on this - I have inspired myself with your PR and incorporated Psr\LoggerInterface into major namespaces rewrite in `namespaces` branch - namely in commits c022384838e7c31076cb147b61fa3a06618e7c27, 4785d5de822d09fd7a9c6b60e5754e8848e7dbb6 and 422622a5e9719b4602be2495dffc56df6e1bf820. 

The implementation is much lighter - as I have noted, there is no need for logging with Exceptions (the end-user himself/herself should do the logging on `catch`) and there is no need to write custom logging for the repository as end-user can choose from wide variety of implementations, namely Monolog with its many adapters.

I am therefore closing this PR withot merging, anyway, thanks for the idea and research.

PR #124, opened 2016-04-28
Remove unused array_insert()

scottconnerly:
Removing array_insert(), as it is unused in this library. Additionally, it is not name spaced and is not prefixed (and is therefore causing conflicts with another package).


finwe:
Merged, thanks!

PR #133, opened 2016-05-17
Fix undefined offset notice and ignore iconv(): Wrong charset.

rayannrag:


finwe:
Can you please provide a code snippet with which those errors show?

rayannrag:
For me, to fix these notices was important because my function returns a json to a jquery ajax. 

screen shot 2016-05-18 at 8 35 21 am
screen shot 2016-05-18 at 8 36 10 am

finwe:
I meant a HTML snippet.

rayannrag:
How I could Do it? My code is in a cakePHP framework.

finwe:
Save contents of `$html` variable to file with file_put_contents and attach that file.

What is your mPDF version?

By the way, are you defining `$this->charset_in` somewhere in config.php? You are most likely setting it to `'ISO-8859-1>'` with the extra closing bracket, which causes the error.

rayannrag:
I am using v6.1.0
The charset is defined in the html, with your tip I fixed the charset error.
This html code comes from [here](https://github.com/CobreGratis/boletophp/blob/master/include/layout_itau.php).
The undefined offset notice still remains.
Thanks.
[html.txt](https://github.com/mpdf/mpdf/files/270522/html.txt)

finwe:
The errors are caused by missing <ul>/<ol> around <li> tags in your HTML. 

Anyway, this PR looks way more as a hack than a solution, so I will not incorporate it in the codebase. More meaningful error messages with invalid HTML are a potential enhancement for the future, when the codebase is cleaned up.

PR #220, opened 2016-08-25
Bugfix: undefined variable in DirectWrite::Write

drjayvee:
The variable `$fill` was always undefined (see [e8dcfd34])

DirectWrite::Write now has a new parameter `$fill`, analogous to other
functions.
`Mpdf::Write` (which delegates to `DirectWrite::Write`) has the same.


finwe:
Thanks! Merged.

PR #228, opened 2016-09-10
Allow any PHP version on Travis

jayvdb:


finwe:
No, sorry. I want to allow only versions against which this was properly tested and is guaranteed to be working.

jayvdb:
The build is failing because `latest` is being built.
I guess there is some flag to override the composer.json php restrictions.

finwe:
Oh, I see. I'll try to take a look.

finwe:
Nice catch with that --ignore-platform-reqs, But I'd allow failures for hhvm and nightly.

jayvdb:
I can understand allowing failure on `nightly`, as it is a moving target, however I do not agree wrt hhvm, which is always the latest release on Ubuntu precise, which is not a new version.  It is 3.6.* which is very stable.

finwe:
It's just that mPDF is not regularly tested on HVVM and not guaranteed to be compatible with it. I know that it most likely will be, but for the time being, at least untill 7.0 is stable and out, I'd like to not to worry that failing hvvm build will distract the development away from more urging matters.

jayvdb:
Done

finwe:
Merged, thanks!

PR #229, opened 2016-09-10
Lint

jayvdb:


finwe:
Thanks for the PR. 

Have you a reasoning why you chose PEAR in particular? What is the result of the CodeSniffer with this configuration? How many style errors?

jayvdb:
Results at https://travis-ci.org/jayvdb/mpdf/jobs/159013533
Happy to switch to a different preset ruleset. options are: Zend, PEAR, PHPCS, Squiz and MySource

jakejackson1:
I'd recommend setting the linter to PSR2

finwe:
:+1: for PSR2 (it is bundled with `phpcs`), however with overruled indentation from spaces to tabs.

Also please remove changes to PHP version constraint (details in #228) and use `^2.7.0` constraint for squizlabs/codesniffer package.

Why is the `- env: LINT=1` line under `allow-failures`?

Thanks!

jayvdb:
Why `^2.7.0` ? That prevents bug fixes from being automatically adopted?

Jobs with `LINT=1` need to be failures, if you want a green build, because the codebase does not pass the phpcs PSR2 (or PEAR) rules.

Here is what a PSR2 job looks like:
https://travis-ci.org/jayvdb/mpdf/jobs/159233138

IMO it is easier to merge it allowed to fail now, fix all the big easy problems (which can be done using `phpcbf` without much assistance), and then tune the linter (which may require raising bugs in the linter and waiting for a new release) , and eventually turning on the lint job as mandatory success much later.

finwe:
`^2.7.0` basically means `<3.0` which by semver means to allow all non-breaking changes. As all other contstrains use `^` notation, I'd like to follow it in this case also.

I understand the need for linter to be allowed to fail, but won't it allow to pass failed builds with errors in tests? I'm not sure, just asking.

I solve passing of builds elsewhere by passing `--runtime-set ignore_warnings_on_exit true` to the phpcs binary, in this case this could be done with `ignore_errors_on_exit` also.

jayvdb:
Please dont merge this yet!  I think using the `script:` block here is incorrect.

We want the `phpcs` job to be red, so people look at it. ;-)
But the `LINT` job shouldn't cause the entire build to be red.

Only jobs with LINT=1 set are allowed to fail, which is only the `phpcs` job.

jayvdb:
Now waiting for feedback on #228 , as this change is less confusing if the Travis builds are green.

jayvdb:
ping

finwe:
Merged, thanks.

PR #231, opened 2016-09-12
Add Travis testing of hhvm-3.3 and hhvm-nightly

jayvdb:
Replace unclear 'hhvm' with 'hhvm-3.6',
and add 'hhvm-3.3' and 'hhvm-nightly' jobs,
also as allowed failures.


finwe:
Merged, thanks!

PR #254, opened 2016-10-29
Correction of the form of instantiation

alanfm:
Correction of the form of instantiation of the FPDF class.


finwe:
Nope. Read the notice at the beginning of the Readme.

PR #315, opened 2017-01-29
SSL improvements and enhancements

KhArtNJava:
1) Added curl_setopt($ch, CURLOPT_FOLLOWLOCATION, 1); to follow HTTP 3xx redirects . This can be useful, when site uses CDN with redirects.

2) Added option $allowUnsafeSslRequests to allow unsafe SSL HTTPS requests. Can be useful when using CDN with HTTPS and if you don't want to configure settings with SSL certificates.


finwe:
Thanks, I've enhanced your PR a bit.

KhArtNJava:
Nice! Thank you for your time!

PR #343, opened 2017-03-07
"fontdata" key added into "defaults" array - otherwise, the variable is overwritten in the constructor.

Azizoff:
Constructor example:
```
$mPDF = new \Mpdf\Mpdf([
    'fontDir' => [
        __DIR__ . '/mpdf/ttfonts/',
        __DIR__ . '/vendor/mpdf/mpdf/ttfonts/'
    ],
    'fontdata' => [
        'arial' => [
            'R' => 'arialn.ttf',
            'B' => 'arialnb.ttf',
            'I' => 'arialni.ttf',
            'BI' => 'arialnbi.ttf',
        ]
    ]
]);
```

### DIFF
--- a/src/Config/ConfigVariables.php
+++ b/src/Config/ConfigVariables.php
@@ -14,7 +14,7 @@ class ConfigVariables
 	{
 
 		$this->defaults = [
-
+            'fontdata' => [],
 			// PAGING
 			'mirrorMargins' => 0,
 			'forcePortraitMargins' => false,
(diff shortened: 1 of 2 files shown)


finwe:
Thanks for the PR. Can you please provide more information and/or reevaluate the change?

The PR should be accompanied with a unit test that fails before the change and passes after it. Also, the whole Travis build must pass in order to th PR to be incorporated.

And lastly, please keep the code style of the library, indentation is made with tabs.

finwe:
Thanks for your effort in moving the library forward. A few more remarks:

- Keep your PRs small, ideally solving one topic/problem
- I will not accept the `var`->`public` change yet. My goal is to properly sort properties and change them accordingly and directly to `public`/`private`. The `var` keyword reminds me this has to be done.
- The `stristr`->`strpos` change will not work for uppercase properties, therefore it is not acceptable as is.

PR #351, opened 2017-03-09
updated functions.php - adding checks if dd() function already exists.

pure4499:
adding checks if dd() function already exists.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -228,12 +228,14 @@ if (!function_exists('cmp')) {
 	}
 }
 
-function dd(...$args)
-{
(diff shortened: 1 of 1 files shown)


finwe:
Thanks!

budirec:
Yes, I don't want to use dev  master, problem is a third-party that I use using dev master

finwe:
They're a bad third party! :) Did you try to talk them out of it?

budirec:
Could be a bad third-party, it's pretty much dead I think. That's why I decide to comment here :) 
Thanks for the quick update though

PR #391, opened 2017-05-08
Fixed headings containing umlaut (ÄÖÜ) not showing up via h2toc

123batman:
I don't know if this is a hack, but I could not find a better place to fix it. 
```
$html = 
'<h2>Chickens</h2>
About chickens...
<h2>Hähnchen</h2>
Über Hähnchen...'
```

Both sections show up fine in the body of the PDF, but only "Chickens" shows up in the TOC. "Hähnchen" generates an empty line only showing the dots and the page number.

Example TOC before:
```
Chickens......1
......1
```
Example TOC after:
```
Chickens......1
Hähnchen......1
```

### DIFF
--- a/src/Tag.php
+++ b/src/Tag.php
@@ -5678,7 +5678,7 @@ class Tag
 						$objattr = [];
 						$objattr['type'] = 'toc';
 						$objattr['toclevel'] = $this->mpdf->h2toc[$tag];
-						$objattr['CONTENT'] = htmlspecialchars($content);
+						$objattr['CONTENT'] = htmlspecialchars(utf8_encode($content));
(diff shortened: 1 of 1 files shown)


finwe:
Thanks for the PR!

Sadly, this doesn't work as expected - when I run example 14 with your code, TOC and bookmarks (I've added utf8_encode to the h2bookmark code part as well) are generated correctly, but the headings are broken.

obrazek

This surely needs more attention, I've created an issue https://github.com/mpdf/mpdf/issues/439

finwe:
Closing as stale and not addressing the issue properly.

PR #401, opened 2017-05-15
Add configuration for hyphenation of web addresses

HKandulla:


HKandulla:
Hi, would be great if you could let me know if you will merge this PR or whether you have some objections? Thanks

PR #402, opened 2017-05-15
Check if the varaible dom is set because in some case this return a notice

marcteyssier:
The case that fix is when the user has
border-top: 1px solid black;
border-right: 1px solid black;
border-top: 2px solid black;
border-bottom: 3px solid black;

On the same tag.
This element fix the problem of notice.


finwe:
Hey, thanks for the pull and sorry for the long wait.

Can you please add a unit test inspired by tests in `tests/issues` directory of last commits on the `development` branch? One that fails/throws a warning/notice without the change and passes with it?

finwe:
This was solved in 7e6ef23, closing.

PR #412, opened 2017-05-31
Fix some 'Notice: Undefined index' for PHP 7.1 compatibility

marinovdf:


src/CssManager.php
```diff
@@ -950,7 +950,7 @@ function setCSSboxshadow($v)
 					}
 				}
 			}
-			if (!$new['col']) {
+			if (!isset($new['col'])) {
```
finwe: This one actually has a different meaning. when `$new['col']` is an empty string, it is set, but would pass in the updated case... Using `empty` or `!empty` would be better.

marinovdf: Fixed in the next commit

src/CssManager.php
```diff
@@ -950,7 +950,7 @@ function setCSSboxshadow($v)
 					}
 				}
 			}
-			if (!isset($new['col'])) {
+			if (!empty($new['col'])) {
```
finwe: It should be `empty`, not `!empty`. And what about other `isset` occurences in your first commit? The same should be applied to them as well (sorry if it weren't clear from the line-comment).

marinovdf: Oh, sorry. Its automatically error. I'll fix it in the next commit (other `isset` too).

marinovdf: Done

marinovdf:
Tests failed. Now I'll look at the code

finwe:
PHP Parse error:  syntax error, unexpected ')' in /home/travis/build/mpdf/mpdf/src/Mpdf.php on line 22081

marinovdf:
There are many potential errors with an undefined index. Maybe it makes sense define all indexes before init? And do you have a code style guide?

finwe:
There are so many places in code I guess it is better to add fixes one by one when failing HTML/CSS snippet is found then to try fixing all at once.

Can you please squash all these commits into one and force push to the same branch? thanks.

finwe:
Thanks!

PR #414, opened 2017-06-02
Add clone feature

kudm761:


finwe:
Thanks for the PR, however, I'd recommend something like https://packagist.org/packages/myclabs/deep-copy to clone nested objects - this PR does not seem to cover all edge cases.

PR #441, opened 2017-07-12
Added _unescape() method

JanSlabon:
This method is needed to support importing of PDF documents with FPDI
which uses escaped sequences in string values if the resulting document
should be encrypted.

See https://stackoverflow.com/questions/45052154/pdf-manipulation-images-are-distorted-after-few-consecutive-operations-on-pdf for details.


finwe:
Thanks!

PR #446, opened 2017-07-18
preg_replace() has a limit

dancameron:
This would prevent the error `preg_replace(): Compilation failed: regular expression is too large at offset...` and allow for all CSS to be loaded.


finwe:
What do you mean - leave CSS strings longer than 1024 bytes with unprocessed URLs? What sense does it make? Why not leave the URL processing altogether then? Not that I am saying it is a viable and possible option at this point.

finwe:
Closing as stale.

PR #451, opened 2017-07-25
Fix php 7.1 typehint

JanGalek:


finwe:
Can you comment on this further, please? What is the error with former version? It is intended behaviour to multiply size with milimeters in an inch divided by document dpi.

Your change does not make sense. If size is not defined, it should result in `0`, not `25.4 / $this->dpi`

Furthermore, 6.x branches are now frozen except for security fixes. All PHP7+ compatibility should be proposed against `development` branch.

JanGalek:
When I use strict types, than I can error message, need float / int but get string.

finwe:
mPDF 6.x is not and will not be PHP 7.1 compatible. You can use the mpdf/mpfg 7.0-dev version for PHP 7.1 compatibility until stable 7.0 is out.

I still do not follow how a type mismatch inside a non-strictly typed file and non static-type-hinted file should influence your strictly typed application. If you are getting a PHP TypeError from an internal PHP function, see the first paragraph.

Issue #457, opened 2017-08-04
[PHP 7.1] Another numeric value error


sebastianblum:
### I have this problem/would like to have this functionality
With our PDF files, we get this error.

Warning: A non-numeric value encountered in /Users/xxx/PhpstormProjects/xxx/vendor/mpdf/mpdf/mpdf.php on line 25843

Warning: A non-numeric value encountered in /Users/xxx/PhpstormProjects/xxx/vendor/mpdf/mpdf/mpdf.php on line 25848

### These are mPDF and PHP versions I am using
7.0.0-beta2
dev-master (f3bbfc7133aed4b7bbdc22b243883bf552558211)

### This is a PHP code snippet I use

```
<?php

require __DIR__.'/../vendor/autoload.php';

$pdfDir = '/Users/xxx/Downloads/pdfs';

$pdfFilePaths = glob(sprintf('%s/*.pdf', $pdfDir));

$mpdf = new  \Mpdf\Mpdf();
$mpdf->debug = true;
$mpdf->SetImportUse();

foreach ($pdfFilePaths as $pdfFilePath) {
    $pagecount = $mpdf->SetSourceFile($pdfFilePath);

    for($i=0; $i < $pagecount; $i++){
        $mpdf->AddPage();
        $template = $mpdf->importPage($i+1, $pdfFilePath);
        $mpdf->useTemplate($template);
    }
}

$mpdf->Output();
```

### This is a HTML code snippet I use
no HTML files

We open existing PDF-files with 1 or several pages and merge them together to one pdf file.

### Solution
I will create a pull request later with unit-tests.

sebastianblum:
here was 1 error:

1) Issues\Issue457Test::testMergePdfFilesWithPHP71
A non-numeric value encountered

/Users/sebiprivat/PhpstormProjects/mpdf/src/Mpdf.php:24416
/Users/sebiprivat/PhpstormProjects/mpdf/src/Mpdf.php:24882
/Users/sebiprivat/PhpstormProjects/mpdf/src/Mpdf.php:11236
/Users/sebiprivat/PhpstormProjects/mpdf/src/Mpdf.php:1976
/Users/sebiprivat/PhpstormProjects/mpdf/src/Mpdf.php:9119
/Users/sebiprivat/PhpstormProjects/mpdf/tests/Issues/Issue457Test.php:21


I can reproduce it with an unit-test now and will now work on the pull request.

JanSlabon:
[...]

PR #458, opened 2017-08-04
Issue #457 Fix for Warning: A non-numeric value encountered

sebastianblum:
Pull request for issue #457 

Please verify, that I understood the code correct:
I assumed that -$tpl['box']['y'] means -1 *$tpl['box']['y']
Is this correct?

Like other forks, I used floatval() to solve the problem.


sebastianblum:
hello @finwe,

do you agree with my solution? Should I change anything?

greetings, Sebastian

finwe:
I guess it's OK. Thanks!

PR #460, opened 2017-08-08
Improve customizability by injecting dependencies

micgro42:
Fixes the strong coupling in the Mpdf constructor. This enables us to, for example, replace Mpdf classes with decorators. E.g. this could be used to decorate the ImageProcessor class and replace links pointing at a web-cache of the local server (/fetch.php?id=...) with a path pointing in the local filesystem (e.g. /var/www/images/...).

### DIFF
--- a/src/Config/ConfigVariables.php
+++ b/src/Config/ConfigVariables.php
@@ -2,17 +2,37 @@
 
 namespace Mpdf\Config;
 
+use Mpdf\Cache;
+use Mpdf\Color\ColorConverter;
+use Mpdf\Color\ColorModeConverter;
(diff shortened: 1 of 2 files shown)


finwe:
This is a good idea in its base, but the execution is somewhat unstable. What if someone forgets a class in overriden configuration? What ensures that the alternative class "implementations" will have the same parameters as their original counterparts? 

I'd also have comments about code style, but as this is in no shape to be incorporated, I'll leave it be.

The dependencies should be really injected via constructor and the defaults should be created with a factory, but that would be much of a BC break in usage of the class for single major version, so I leave this for later. This can be done right, but I'd like it to be designed right and foolproof.

PR #465, opened 2017-08-10
compatibility with php 7.1 for the 6.1 branch

Linguisto:
Fixed "A non-numeric value encountered" warning in the "ConvertSize" function that appears on php7.1 with 6.1.* versions of the package.


finwe:
6.x branch is frozen and will not offer PHP 7.1 compatibility. 

There are far more places in the code that do not work properly in PHP 7.1. Most of them are fixed in 7.0-dev branch.

PR #490, opened 2017-09-19
protect against Divide By Zero error

cliffordp:
Avoid possibility of dividing by zero

With these:
```
$mpdf = new mPDF(
	'UTF-8', //
	'LETTER', // default is A4
	'12' // I think default is 9
);
```

I kept hitting Divide By Zero error only at this line.

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -21408,8 +21408,10 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 						}
 					} elseif ($numcols) { // If all columns
 						$ttl = array_sum($table['l']);
-						for ($i = 0; $i < $numcols; $i++) {
-							$widthcols[$i]['miw'] += $surplus * $table['l'][$i] / $ttl;
+						if ( ! empty( $ttl ) ) {
+							for ( $i = 0; $i < $numcols; $i ++ ) {
+								$widthcols[ $i ]['miw'] += $surplus * $table['l'][ $i ] / $ttl;
+							}
 						}
 					}
 				}


finwe:
Thanks for the PR. Can you add a unit test failing before and passing after your change? See `tests/Issues` directory in the repo. Also, see and fix Travis build failures.

cliffordp:
@finwe, thanks for being willing to consider my pull request. To be honest, my PR is more of a band aid / short circuit fix for the specific divide by zero error I was experiencing. I see in the same section of code other instances that divide by a variable without first checking if the variable is zero.

Therefore, please consider it as flagging the issue of potentially dividing by zero (should first check if the variable is zero or fix the issue of it somehow being zero in the first place) -- especially since I'm running nearly the defaults.

FYI: I'm running [v6.1.4](https://github.com/mpdf/mpdf/releases/tag/v6.1.4) but obviously the issue is still in _development_ branch because I submitted the PR on that branch.

Thank you for taking this into consideration and hopefully being able to _fully_ fix yourself.

Klap-in:
Because it is just one case, the unit test is really helpful and makes visible what is going on.
That makes it easier to expand it to the other cases in a later stage.
(due to the complexity of the code it is mosttimes not doable, and difficult to test, to perform all such improvements at once.)

cliffordp:
Thanks for explaining, @Klap-in 

I would do it quick if I knew how... Do you not get the same error when running as I indicated?

cliffordp:
The error happened for me in mPDF v7.0.1 and the same fix worked for me in PHP 5.6.2 running mPDF 7.0.1, this time running it as `new \Mpdf\Mpdf( array( 'c' ) );`

cliffordp:
Just confirming that this consistently happens with v7.0.2 with the same test HTML file shared at https://github.com/mpdf/mpdf/issues/524#issuecomment-340215521

PR #491, opened 2017-09-20
fixed colsum float calculation for german based locales

flow-control:
In germany we use comma as decimal point and dot as thousands separator.  `floatval()` is not locale aware (as is `scanf()`), so we must use `str_replace()` magic :wink:

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -23929,17 +23929,10 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 									$cell['textbuffer'][0][0] = preg_replace('/{colsum[0-9_]*}/', $rep, $cell['textbuffer'][0][0]);
 								}
 							} elseif (!isset($table['is_thead'][$i])) {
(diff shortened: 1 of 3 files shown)


finwe:
Thanks for a PR. I am hesitant to incorporate this as this could break other languages and/or formats. It is kind of hacky for one particular use-case.

flow-control:
Damn, you are right. I thought i took care of this, but it won't work other locales ... My fault. I will build a more general solution for this problem and update this PR

finwe:
Can you also add a unit test with HTML/CSS code demonstrating your change? See `tests/Issues` directory in the repo. Also, see and fix Travis build failures.

It would be better to create a mPDF method instead of a function with such generic name.

flow-control:
Moved the function, i will now create the unit test

flow-control:
Hi @finwe, Unittests are pushed and are green, do you have any further hints or notes on this PR?

finwe:
Merged, thanks a lot!

PR #492, opened 2017-09-20
fixed colsum float calculation for german based locales (backport)

flow-control:
In germany we use comma as decimal point and dot as thousands separator.  `floatval()` is not locale aware (as is `scanf()`), so we must use `str_replace()` magic :wink: 

This is a backport from #491 to the current stable version 6.1.4


finwe:
Please note that 6.x branch is frozen. I will not incorporate this.

PR #499, opened 2017-10-03
Use constant instead of expression when define SCALE const

ImmRanneft:
Define constant as float, not numeric. This is only thing stops to use v7 with php 5.5 I encountered.


finwe:
PHP 5.5 is not a supported version for 7.x branch, sorry. 

It is also after its end of life and does not receive any support. You should upgrade at least to 5.6 as soon as possible.

See https://secure.php.net/supported-versions.php

PR #513, opened 2017-10-18
Update TTFontFile.php

kravcik:
Some fonts (ex. http://font.ubuntu.com/) throws notice of undefined index. Simple fix.

image


finwe:
Cool, thanks!

PR #515, opened 2017-10-20
Correct release date

marclaporte:


finwe:
OMG. So much time on this version and then so many errors made in haste. Thanks, will correct directly to development branch.

PR #518, opened 2017-10-23
Custom color for QRCode

achretien:
Allow a color customization for the QRCode

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -7217,13 +7217,32 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 						$this->qrcode->disableBorder();
 					}
 
+					$bgColor = [255, 255, 255];
+					if ($objattr['bgcolor']) {
+						$bgColor = array_map(
+							function ($col) {
+								return intval(255 * floatval($col));
+							},
+							explode(" ", $this->SetColor($objattr['bgcolor'], 'CodeOnly'))
+						);
+					}
+					$color = [0, 0, 0];
+					if ($objattr['color']) {
+						$color = array_map(
+							function ($col) {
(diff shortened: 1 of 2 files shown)


finwe:
Thanks for the PR. However, this will not work unless you also assign $objattr keys in Tag class. Also, see failing Travis build and fix errors. And write a test class using your new feature to keep track of the HTML code and assure that the code does not cause any notices and/or warnings. See tests/Issues folder.

Amend the changes and force push to this PR. Thanks!

achretien:
@finwe How may I test the color of my QRCode ?

finwe:
I generally use tests without assertions just to check for hidden warnings or notices.

You can see Issue442Test.php - if you write a simple document twice with QR code of different colors, the output should differ. In the uncompressed PDF, you can find the specification of the color based on where the two strings differ and assert that the color definition is present.

achretien:
Ok just added the test

mindeffects:
I cannot see any real use case for QR-codes in a PDF on screen (aka RGB), since QR-codes are just fancy talk for "link" ;-) . They only make sense in print or "on something real", so support of CMYK- and spot-colors would come in more than handy!

To get CMYK-QRs I used to hack these lines in "QrCode.php":
```
148: $mpdf->SetDrawColor(0,0,0,100); // Border
149: $mpdf->SetFillColor(0,0,0,0);   // Background
162: $mpdf->SetFillColor(0,0,0,100); // Foreground
```
Works like a charm! Even with PDF/X-1a! Would be great to have that in mPDF 7.1!
Did I mention that mPDF just rocks? Well, it does!!!

finwe:
Merged in 8c271ce. Feel free to open a new Issue with CMYK/spot colors for QRCodes.

PR #519, opened 2017-10-23
Better SVG path parsing

achretien:
Better parsing of SVG files. (Rebase of an old PR done on 6.1.3)


finwe:
Thanks for the PR. I doubt that with removing the [0] keys from the code would maintain current functionality and not cause a BC break.

Can you please elaborate what are the errors this should fix? What are the possible side effects?

Also, a test class would be nice, as well as a passing Travis build.

achretien:
The regex has been updated to handle more SVG possibilities.
I added a test with an SVG which fails on current dev and works with mine. 

I added a comment on top of the regex with a link to the source of the regex.

achretien:
@finwe could you check the PR?

PR #528, opened 2017-10-31
isset in CssManager.php

adam2314:
$bg array is not always defined and not always with a p cell
this should be better


Klap-in:
Do you have also a minimal test case that reproduces your case? Could you add it please as unit test? For examples please see https://github.com/mpdf/mpdf/tree/development/tests
This guarantees that future developments will not break it again.

finwe:
Solved in 33fb44c, closing.

Issue #502, opened 2017-10-06
Public members in \Mpdf\Mpdf


KacerCZ:
Class `\Mpdf\Mpdf` contains huge amount of public members.
Properties are declared using `var` and methods don't have accessibility specified making them effectively `public`. Some of methods have name starting with underscore probably indicating that they are meant to be `private`.

Release of new major version is good opportunity to correct this, but given amount of changes already present in 7.0 this seems it should be done in 8.0.

Is there any plan to change this and clean-up class interface?

finwe:
Yes, it should be done. However triaging the properties turned out to be very hard.

I have a vague plan how to do this (creating a few value objects that would be shared among service classes, some public, some private), but I have not yet come up with a way to do this properly.

There are more things to consider - especially changes to the current "settings" approach via direct property modification.

With this change I would also like to unify the properties to a single camelCase format.

Much work to be done there.

PR #542, opened 2017-11-18
PHP oop syntax fixes

machour:
This PR:
  * Replaces old `var` properties declaration with `public`
  * Adds `public` to methods without access modifier

I switched everything to public to avoid any BC breaks, separated commits & successfully ran tests.

### DIFF
--- a/src/CssManager.php
+++ b/src/CssManager.php
@@ -31,21 +31,21 @@ class CssManager
 	 */
 	private $colorConverter;
 
-	var $tablecascadeCSS;
+	public $tablecascadeCSS;
 
-	var $cascadeCSS;
+	public $cascadeCSS;
 
-	var $CSS;
+	public $CSS;
 
-	var $tbCSSlvl;
+	public $tbCSSlvl;
 
-	var $cell_border_dominance_B;
+	public $cell_border_dominance_B;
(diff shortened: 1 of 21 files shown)


Klap-in:
Not sure, but it could be that the `var` instead of `public` property was used for some bookkeeping, with respect to future refactorizations.

machour:
@Klap-in thanks for this insight, haven't thought of it!
Maybe this commit can be used for the book keeping instead: https://github.com/mpdf/mpdf/pull/542/commits/7ab0601c161f0b9ff60365249572b3d1548abd08
😄

finwe:
Thanks, but I'd really like to have the `var` keywords right there in the open to cause pain every time anyone comes across them so that he/she knows that this is still unaccounted for and needs *real* fixing :)

machour:
@finwe I'm curious what would the _real_ fixing look like? Are you talking about getting rid of this properties, or picking `private` or `public` depending on the need?

finwe:
The latter - picking private or public depending on the needs. Also see https://github.com/mpdf/mpdf/issues/502

machour:
Thank you! I'll take a closer look at the project's milestones.

PR #543, opened 2017-11-18
Conversion code clean up

machour:
- Optimisation: Use (int) instead of intval()
- Optimisation: Don't use count() or strlen() in `for` condition
- Optimisation: Use ** (>= 5.6) instead of `pow()`
- Use strict comparison (===) where needed
- Some code readability changes ($c1 initialized with default value, default property value in declaration)


machour:
(I don't understand why this PR is showing discussions & commits from an old PR)

machour:
Was branching from master instead of development. Fixed in #546

PR #544, opened 2017-11-18
Language code clean up

machour:
- Use single quotes instead of double quotes
- Use static properties for heavy declarations
- Remove unnecessary regex for splitting
- Use !empty($foo) instead of isset($foo) && $foo
- Strict type checking for comparisons
- Refactor ifs to shorten them


machour:
Sorry, I was branching for master instead of development -_-, fixed in #545

PR #545, opened 2017-11-18
Language code cleanup

machour:
- Use single quotes instead of double quotes
- Use static properties for heavy declarations
- Remove unnecessary regex for splitting
- Use !empty($foo) instead of isset($foo) && $foo
- Strict type checking for comparisons
- Refactor ifs to shorten them

(Replaces #544)

### DIFF
--- a/src/Language/ScriptToLanguage.php
+++ b/src/Language/ScriptToLanguage.php
@@ -7,7 +7,7 @@ use Mpdf\Ucdn;
 class ScriptToLanguage implements \Mpdf\Language\ScriptToLanguageInterface
 {
 
-	private static $scriptDelimiterMap = [
+	private $scriptDelimiterMap = [
 		'viet' => "\x{01A0}\x{01A1}\x{01AF}\x{01B0}\x{1EA0}-\x{1EF1}",
 		'persian' => "\x{067E}\x{0686}\x{0698}\x{06AF}",
 		'urdu' => "\x{0679}\x{0688}\x{0691}\x{06BA}\x{06BE}\x{06C1}\x{06D2}",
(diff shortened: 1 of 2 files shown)


Klap-in:
Out of curiosity a question: why to use single quotes instead of double quotes?

machour:
@Klap-in content inside of "" is analyzed by the PHP engine to substitute $variables. Content inside of '' is not. This [SO entry](https://stackoverflow.com/a/3446286/732772) may give you more information. 
It's a common used coding standard in PHP to use single quotes when possible.

src/Language/ScriptToLanguage.php
```diff
@@ -7,15 +7,15 @@
...
+	private static $scriptDelimiterMap = [
 		'viet' => "\x{01A0}\x{01A1}\x{01AF}\x{01B0}\x{1EA0}-\x{1EF1}",
 		'persian' => "\x{067E}\x{0686}\x{0698}\x{06AF}",
 		'urdu' => "\x{0679}\x{0688}\x{0691}\x{06BA}\x{06BE}\x{06C1}\x{06D2}",
 		'pashto' => "\x{067C}\x{0681}\x{0685}\x{0689}\x{0693}\x{0696}\x{069A}\x{06BC}\x{06D0}", // ? and U+06AB, U+06CD
 		'sindhi' => "\x{067A}\x{067B}\x{067D}\x{067F}\x{0680}\x{0684}\x{068D}\x{068A}\x{068F}\x{068C}\x{0687}\x{0683}\x{0699}\x{06AA}\x{06A6}\x{06BB}\x{06B1}\x{06B3}",
 	];
 
-	private $scriptToLanguageMap = [
+	private static $scriptToLanguageMap = [
```
finwe: Why static? Any particular reason for this? I think static should be avoided if possible, it is private so there are no benefits for this in class' non-static methods.

machour: It was a _very_ tiny memory optimisation while I was there. 
In fact, I wanted to turn this into a const at first, but private const are not available before PHP 7.1.
I'll rollback this change

machour:
@finwe should be good to go now.

PR #546, opened 2017-11-18
Conversion code cleanup

machour:
- Optimisation: Use (int) instead of intval()
- Optimisation: Don't use count() or strlen() in `for` condition
- Optimisation: Use ** (>= 5.6) instead of `pow()`
- Use strict comparison (===) where needed
- Some code readability changes ($c1 initialized with default value, default property value in declaration)

Replaces #543

### DIFF
--- a/src/Conversion/DecToCjk.php
+++ b/src/Conversion/DecToCjk.php
@@ -14,7 +14,7 @@ class DecToCjk
 		$glyphs = [0x3007, 0x4E00, 0x4E8C, 0x4E09, 0x56DB, 0x4E94, 0x516D, 0x4E03, 0x516B, 0x4E5D];
 		$len = strlen($nstr);
 		for ($i = 0; $i < $len; $i++) {
-			$rnum .= UtfString::code2utf($glyphs[(int)$nstr[$i]]);
+			$rnum .= UtfString::code2utf($glyphs[(int) $nstr[$i]]);
 		}
 		return $rnum;
 	}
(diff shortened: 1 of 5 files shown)


src/Conversion/DecToCjk.php
```diff
@@ -12,8 +12,9 @@ public function convert($num)
 		$nstr = (string) $num;
 		$rnum = '';
 		$glyphs = [0x3007, 0x4E00, 0x4E8C, 0x4E09, 0x56DB, 0x4E94, 0x516D, 0x4E03, 0x516B, 0x4E5D];
-		for ($i = 0; $i < strlen($nstr); $i++) {
-			$rnum .= UtfString::code2utf($glyphs[intval($nstr[$i])]);
+		$len = strlen($nstr);
+		for ($i = 0; $i < $len; $i++) {
+			$rnum .= UtfString::code2utf($glyphs[(int)$nstr[$i]]);
```
finwe: How about keeping a space everywhere after type casts?

src/Conversion/DecToRoman.php
```diff
@@ -49,7 +47,7 @@ private function ensureNumberIsWithinBounds($number)
 	public function getUpperBound()
 	{
 		$symbolGroupCount = count($this->symbolMap);
-		$valueOfOne = pow(10, $symbolGroupCount - 1);
+		$valueOfOne = 10 ** ($symbolGroupCount - 1);
```
finwe: Nice.

src/Conversion/DecToRoman.php
```diff
@@ -9,14 +9,12 @@
...
 
-	private $symbolMap;
+	private $symbolMap = [['I', 'V'], ['X', 'L'], ['C', 'D'], ['M']];
 
 	public function __construct(array $symbolMap = [])
 	{
 		if ($symbolMap !== []) {
 			$this->symbolMap = $symbolMap;
-		} else {
-			$this->symbolMap = [['I', 'V'], ['X', 'L'], ['C', 'D'], ['M']];
```
finwe: I'd really like to have properties initialized in constructors of the class. It offers more variability and keeps potential logic in one place (as like exactly in this case).

machour:
Requested changes performed!

PR #553, opened 2017-11-20
Suppress iconv notice

VolCh:
Suppress PHP notice 'iconv(): Detected an illegal character in input string in /var/www/vendor/mpdf/mpdf/src/Mpdf.php:28002' on some strange content.


finwe:
This is not a correct solution for this as it could hide potential harmful errors. Closing this, please file an issue (using the template) with details about when this is happening.

PR #568, opened 2017-11-27
Bmp code cleanup

machour:
Code cleanup & correct access modifiers set (scanned the src/ folder for usage).
Also marked an undefined var with // fixme


finwe:
Great work, thanks!

PR #571, opened 2017-11-29
Fixed bitshift by 7128 bits, must be 7 bits

### LINKED ISSUE #569: Bitshift by 7128 bits?
https://github.com/mpdf/mpdf/blob/v5.4.0/classes/bmp.php#L93
https://github.com/mpdf/mpdf/blob/v7.0.1/src/Image/Bmp.php#L106

Looks like it's supposed to be 7 :)

5lava:
Fixes #569.


finwe:
Good catch, thanks!

PR #572, opened 2017-11-30
Gradient cleanup

machour:
* Classical fixes
* accurate access modifiers (scanned source code)
* split parseMozGradient to use two new private methods
* fixed a `break2` typo

### DIFF
--- a/src/Gradient.php
+++ b/src/Gradient.php
@@ -130,10 +130,10 @@ class Gradient
 	// $stops = array('col'=>$col [, 'opacity'=>0-1] [, 'offset'=>0-1])
 	function Gradient($x, $y, $w, $h, $type, $stops = [], $colorspace = 'RGB', $coords = '', $extend = '', $return = false, $is_mask = false)
 	{
-		if (strtoupper(substr($type, 0, 1)) === 'L') {
+		if (stripos($type, 'L') === 0) {
 			$type = 2;
 		} // linear
-		else if (strtoupper(substr($type, 0, 1)) === 'R') {
+		else if (stripos($type, 'R') === 0) {
 			$type = 3;
 		} // radial
 		if ($colorspace !== 'CMYK' && $colorspace !== 'Gray') {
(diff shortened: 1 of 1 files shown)


src/Gradient.php
```diff
@@ -372,13 +375,9 @@ function Gradient($x, $y, $w, $h, $type, $stops = [], $colorspace = 'RGB', $coor
 				// All values are set in parseMozGradient - so won't appear here
 				$coords = [0, 0, 1, 0]; // default for original linear gradient (L2R)
 			}
-			$s = ' q';
-			$s .= sprintf(' %.3F %.3F %.3F %.3F re W n', $x * Mpdf::SCALE, ($this->mpdf->h - $y) * Mpdf::SCALE, $w * Mpdf::SCALE, -$h * Mpdf::SCALE) . "\n";
-			$s .= sprintf(' %.3F 0 0 %.3F %.3F %.3F cm', $usew * Mpdf::SCALE, $useh * Mpdf::SCALE, $usex * Mpdf::SCALE, ($this->mpdf->h - ($usey + $useh)) * Mpdf::SCALE) . "\n";
 		} // RADIAL
-		else if ($type == 3) {
+		else if ($type == self::TYPE_RADIAL) {
```
finwe: Could we have `elseif` everywhere and on the same line as previous closing brace? With the constant used for type, the comment is unnecessary.

src/Gradient.php
```diff
@@ -22,6 +22,9 @@ class Gradient
 	 */
 	private $colorConverter;
 
+	const TYPE_LINEAR = 2;
```
finwe: Please, keep class constants declared first in the class. The order is constants, properties, constructor, other methods (preferrably public first, private second).

src/Gradient.php
```diff
@@ -587,370 +585,372 @@ function Gradient($x, $y, $w, $h, $type, $stops = [], $colorspace = 'RGB', $coor
...
+				if ($tmp) {
+					$starty = $m[1];
 				}
 			}
 
-			$coords = [];
-			if (!isset($startx)) {
-				$startx = false;
-			}
+			/*
```
finwe: Maybe this whole part could be removed?

machour:
All done!

finwe:
Cool. I still see many redundant comments above else if blocks. Could you please
- change `else if` to `elseif` everywhere
- move the elseif on the same lines as previous `}`
- if the comment above is redundant, delete it, if not, move it after the elseif condition
?
Thanks a lot!

machour:
Done. Also did a little more refactoring to avoid redundancies

finwe:
This is fine, but changes to the `parseMozGradient` method technically make this still a BC break (although it is probable noone ever used that method anywhere.

Would it be possible to add a public `parseMozGradient` method reusing those two private methods and mark it as deprecated so it could be removed in next major version?

machour:
It's still there and needed: https://github.com/mpdf/mpdf/pull/572/files#diff-4c7dc10f785e5e8c0016dc7ab150cda6R899

machour:
Should be good to go too.

PR #573, opened 2017-12-01
Colorconverter cleanup

machour:
* Remove unnecessary parentheses
* Use single quotes
* Use short syntax
* Strict comparison for strings
* Clean up ifs
* Add missing phpdoc annotations

### DIFF
--- a/src/Color/ColorConverter.php
+++ b/src/Color/ColorConverter.php
@@ -39,7 +39,6 @@ class ColorConverter
 	public function convert($color, array &$PDFAXwarnings = [])
 	{
 		$color = strtolower(trim($color));
-		$cstr = '';
 
 		if ($color === 'transparent' || $color === 'inherit') {
 			return false;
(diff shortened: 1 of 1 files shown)


src/Color/ColorConverter.php
```diff
@@ -199,8 +213,8 @@ private function processHashColor($color)
 	/**
 	 * @param $mode
 	 * @param mixed[] $cores
-	 *
-	 * @return bool|\float[]
+	 * @return bool|float[]
+	 * @throws \Mpdf\MpdfException
```
finwe: I discourage `@throws` PHPDoc tag, it is a mess to maintain, better to ommit it.

Klap-in: Could it be useful if automatically docs are generated from the source?

finwe: IMHO not too much. There is not enough difference in exceptions thrown (now there are about 2 or 3 types) and when the exception change in code is not reflected in the phpdoc, the generated doc would be off as well.

machour: I agree it's a pity to maintain, reverted the change :)

PR #582, opened 2017-12-13
Appended missing Class for static method in Tag.php

anber500:
Fix two places where code2utf was called without it's Class appended. This was causing some renders to break.


finwe:
Thanks!

PR #604, opened 2018-01-09
Do not render table backgrounds when considering page breaks.

### LINKED ISSUE #570: Page break inside: 'avoid' can cause table background to be copied to previous page
### I found this bug / would like to have this new functionality

When you have markup like this:

```
<table>...</table>

<div style="page-break-inside:avoid">
    <table bgcolor="red">...</table>
</div>
```

...and mPDF decides to move the page break div to a new page, the previous page will contain the background color of the table which was moved to the new page.

The first table is required for this bug to present itself.

### This is mPDF and PHP version and environment (fpm/cli etc) I am using

mPDF 7.0.2
PHP 7.0.25

### This is a PHP code snippet I use

```
<?php
require 'vendor/autoload.php';

$mpdf = new Mpdf\Mpdf;

$before = str_repeat('<p>Before</p>', 20);
$after = str_repeat('<p>After</p>', 10);

$html = <<<HTML
    $before

    <table>
        <tr>
            <td>Normal table cell</td>
        </tr>
    </table>

    <div style="page-break-inside:avoid">
        <table>
            <tr>
                <td bgcolor="red">Red table cell</td>
            </tr>
        </table>

        $after
    </div>
HTML;

$mpdf->writeHtml($html);
$mpdf->output();
```

### Example of incorrect rendering

image

RyanDwyer:
Fixes #570.

Wrapping the background printing code in a $this->keep_block_together check solves my issue.

This is my first patch to mPDF. I highly recommend reviewing my change before merging, as I'm not familiar with the mPDF codebase and I'm not confident that I haven't broken anything. I haven't broken any unit tests at least.

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -22941,21 +22941,23 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 		}
 
 		if ($this->tableBackgrounds && $level == 1) {
-			$s = $this->PrintTableBackgrounds();
-			if ($this->table_rotate && !$this->processingHeader && !$this->processingFooter) {
-				$this->tablebuffer = preg_replace('/(___TABLE___BACKGROUNDS' . $this->uniqstr . ')/', '\\1' . "\n" . $s . "\n", $this->tablebuffer);
-				if ($level == 1) {
-					$this->tablebuffer = preg_replace('/(___TABLE___BACKGROUNDS' . $this->uniqstr . ')/', " ", $this->tablebuffer);
-				}
-			} elseif ($this->bufferoutput) {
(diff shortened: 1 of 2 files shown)


finwe:
Thanks for the PR! Will look into it, hopefully in a few days.

eleva:
Hi there, I'm having a problem with the "page-break-inside: avoid" property on a div. It adds a break page correctly, but the backgrounds of the elements before the break page stop working. I try to add the background-color inline and through CSS but nothing works. I couldn't find any issue about it either. Can you guys help us out? Thanks

RyanDwyer:
I've rebased this. Please review.

PR #612, opened 2018-01-15
fix warning line 17074 - a non-numeric value encountered

wideti:
I changed these lines because I was using the symfony framework, and when I run the project, the server accused the error "a non-numeric value encountered", because in the element it received a string instead of a numeric value.


finwe:
Upgrade to mPDF 7.x, where this is already accounted for. 6.x branch is frozen except for serious security fixes.

PR #614, opened 2018-01-17
Fix throwing PHP Warning. If mpdf import pdf tpl.

PATROMO:
PHP Warning:  Invalid argument supplied for foreach() in mpdf/src/Mpdf.php on line 25272.
Because `$v[1]` is not an array.

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -25268,7 +25268,7 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 						$this->current_parser = $tpl['parser'];
 
 						foreach ($tpl['resources'][1] as $k => $v) {
-							if ($k == '/Shading') {
+							if ($k == '/Shading' && is_array($v[1])) {
 								foreach ($v[1] as $k2 => $v2) {
 									$this->_out($k2 . " ", false);
 									$this->pdf_write_value($v2);


finwe:
Can you please add a test-case which would represent this behaviour?

PATROMO:
@finwe I dont know how I should write a test for this problem. The template pdf which trigger the problem belongs our customer and I want publish this. :(

The pdfi parser return in `getPageResources()` this:

````
array (2)
   0 => 5
   1 => array (6)
   |  "/ProcSet" => array (2)
   |  |  0 => 6
   |  |  1 => array (2) [ ... ]
   |  "/ColorSpace" => array (3)
   |  |  0 => 8
   |  |  1 => 33
   |  |  2 => 0
   |  "/ExtGState" => array (3)
   |  |  0 => 8
   |  |  1 => 34
   |  |  2 => 0
   |  "/Pattern" => array (3)
   |  |  0 => 8
   |  |  1 => 35
   |  |  2 => 0
   |  "/Shading" => array (3)
   |  |  0 => 8
   |  |  1 => 36
   |  |  2 => 0
   |  "/Font" => array (3)
   |  |  0 => 8
   |  |  1 => 37
   |  |  2 => 0
````
 

And the second entry of "/Shading" is not an array.
But Mpdf import our customer template correct.

finwe:
I need a usage example in order to at least be able to reproduce this for myself. Until that, I will not encorporate this.

JanSlabon:
@finewe Is there any PDF and viewer arround which confirms the need of the copy action of the Shading resources in mPDF at all? We never had any bug report in view to this and we never did such things when using FPDI with FPDF or TCPDF. IMHO: This is useless code. Which is also confirmed by the "fix" that is suggested in this pull request: It only silents the warning but the Shading resources are still not assigned/copied to the global resources dictionary.

PR #623, opened 2018-01-29
Improve error message if headers already sent

thomas-dotworx:
This PR adds the filename and the line where the header output started
image


finwe:
Nice improvement, thanks!

Issue #534, opened 2017-11-09
mkdir(): File exists in Cache constructor


aluferraz:
### I found this bug 
Hi,

I'm facing the following error:

"mkdir(): File exists"

It occurs whenever I instantiate a new MPdf object. The error occurs at:

vendor/mpdf/mpdf/src/Cache.php in __construct at line 18

### This is mPDF and PHP version and environment (fpm/cli etc) I am using

mPDF version 7.0.1
PHP Version: 7.1.10
LAMP stack

### This is a PHP code snippet I use

```
<?php

$mpdf = new Mpdf();
```

### This is a HTML/CSS code snippet I use

```

```

finwe:
How did you install mPDF? with composer? There might be an issue with correct permissions to to the temp directory.

I strongly suggest to create your own custom temporary directory with write permissions and set it to mPDF as `$mpdf = new Mpdf(['tempDir' => '<custom temp path>']);`

aluferraz:
Hi there,

Thank you for your reply! MPDF was installed using composer.. I've changed the folder permissions to 777 and the problem was fixed. However, shouldn`t it already 777? I mean, this folder is created by the script itself using mkdir(<tmp dir path>,777,true).

finwe:
Yes, there is a script run after installation, but in a case your `umask` is `0002`, resulting mode is `0775` (no write permissions for `others`). When you run composer as yourself, the Apache which executes PHP script is user `www-data`, there is no mutual group and therefore Apache, as `others`, has no write permissions.

I seriously recommend to use your own custom path with correct permissions, directory owner and group.

Leaving this open until I add a patch throwing an exception instead of a warning in this particular corner case.

finwe:
Just out of curiosity, can you send owner, group and permissions of all directories along path `vendor/mpdf/mpdf/tmp`?

aluferraz:
`$ ls -lha`
drwxrwsr-x  3 user user 4.0K Nov  6 17:31 tmp

finwe:
and a few directories above? I guess that this is exactly the case I described above.

aluferraz:
Yes, I think you're right ! I've changed the tmp folder to a custom location as a fix..

`$ vendor/mpdf/mpdf/ ls -lha`

-rw-rw-r--  1 user user  33K Oct 20 19:45 CHANGELOG.md
-rw-rw-r--  1 user user 1.5K Oct 20 19:45 composer.json
[...]

PR #624, opened 2018-01-29
Fix handling permissions on temporary directories

vpyatnitskiy:
Follow-up for #534. Due to directory permissions issue, it is currently impossible to generate a PDF using one font, then generate another one using a different font and running under a different user (i.e. create a PDF with "font1" from the command line, then another PDF with "font2" from a web server), because the latter cannot write to `tmp/ttfontdata` directory, whenever `umask() !== 0` ― which is usually the case.

**Issue:**

`Mpdf\Cache` uses `mkdir()` to create missing temporary directories.

`mkdir()` takes umask into account (see [PHP documentation](http://php.net/mkdir#refsect1-function.mkdir-parameters)). It's not unusual to have `0002` or `0022` umask by default, so the directories end up not being world-writeable (i.e. having file mode `0775` or `0755` instead of `0777`).

This means that if a CLI worker creates the temp dirs first, a web server (running under a different, unrelated user) may not be able to generate PDF, and vice versa.

Whereas it is possible to create the main temporary directory and assign the correct permissions to it in advance, it is not true for `ttfontdata` subdirectory, which is created by MPDF itself ― so the ability to manually correct file permissions is limited.

**Fix:**

Use `chmod()`, which doesn't use umask, to assign permissions.

Since `chmod()` does not support recursive assignments, it was necessary to implement recursive directory creation manually. In the PR, `Mpdf\Cache::createBasePath()` function recursively calls itself to check the parent directory before attempting to create the target.

**Steps to reproduce:**

1. Initialize a project:
  
  ```
  composer init
  composer require mpdf/mpdf
  ```

2. Put the following code into a PHP script, and put **any two** TTF files in the same directory:

```php
<?php

require "vendor/autoload.php";

error_reporting(E_ALL);
ini_set('display_errors', 'On');

$fonts = glob('*.ttf');
sort($fonts);

$fontIndex = php_sapi_name() === "cli" ? 0 : 1;

$mpdf = new Mpdf\Mpdf([
     'fontDir' => __DIR__,
     'fontdata' => [
        'font-' . $fontIndex => [
                'R' => $fonts[$fontIndex],
        ],
     ],
]);

$mpdf->WriteHtml('<h1 style="font-family: font-' . $fontIndex . ';">Hello World!</h1>');
$mpdf->Output();
```

3. Run the script from the command line (i.e. `php index.php`).

4. Run the same script using a web server. Observe the generated exception.

**Other notes:**

I didn't find any existing tests pertaining to the `Cache` class, so I created a stub and put my tests there. The stub may be later expanded to completely cover the class. However, this is outside of the scope of the PR at the moment. Comments are welcome.

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -13,18 +13,44 @@ class Cache
 
 	public function __construct($basePath, $cleanupInterval = 3600)
 	{
-		if (!is_writable($basePath)
-				&& is_writable(dirname($basePath))
-				&& !file_exists($basePath)) {
-			mkdir($basePath, 0777, true);
-		}
-
-		if (!is_writable($basePath)) {
+		if (!$this->createBasePath($basePath)) {
 			throw new \Mpdf\MpdfException(sprintf('Temporary files directory "%s" is not writable', $basePath));
 		}
 
 		$this->basePath = $basePath;
 		$this->cleanupInterval = $cleanupInterval;
+	}
+
+	protected function createBasePath($basePath)
+	{
+		if (!file_exists($basePath)) {
+			if (!$this->createBasePath(dirname($basePath))) {
+				return false;
+			}
+
+			if (!$this->createDirectory($basePath)) {
+				return false;
+			}
+		}
+
+		if (!is_writable($basePath) || !is_dir($basePath)) {
+			return false;
+		}
+
+		return true;
+	}
+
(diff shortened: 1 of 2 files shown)


finwe:
Awesome. I wish all PRs were as elaborate as this one. Thanks.

PR #627, opened 2018-02-01
Added most popular Bangla Fonts

AnowarCST:
Hello, I just added some popular Bangla/Bengali fonts for default font selection in Bangla Language.


finwe:
Thanks for the PR, but I will close it for now. It is not sustainable for mPDF to contain fonts for every concievable language. In the future, I plan to have separate composer packages with additional fonts, now there is no real point.

You can use custom fonts on the user side easily enough.

Issue #653, opened 2018-02-14
Internal anchors not linking with ids


amnausman:
mPDF don't link anchors based on id attribute, for example <a id="mm"></a> will not work. Although  name attribute is recognised and linked but this should be marked as issue since according to HTML 5 specification， we should use id attribute instead of name for anchors.

dbielke1986:
Would be fine if id tag can be supported with internal anchors, because according to the [HTML 5 specification](https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-fragid), the id is the way to go.

vincenthoehn:
+1

PR #657, opened 2018-02-22
feat: Allow ID to be used as internal anchor

machour:
I had a look at #653 and added support for `id` as anchor in the `A` tag.

One thing is not working though: The value used in `id` gets automatically uppercased in `Mpdf.php` for a CSS thingy. (see added comment)

So this is **not** working:
```html
<a href="#foo">Go to foo</a>
<pagebreak />
<a id="foo">Here I am</a>
```

While this is working as expected:
```html
<a href="#FOO">Go to foo</a>
<pagebreak />
<a id="FOO">Here I am</a>
```

@finwe do you happen to know what problem was that line trying to solve?


finwe:
No idea. Thanks for the PR, will look into it probably tomorrow.

jakejackson1:
@machour can you add unit tests to this PR? Does it work if the ID `FOO` is assigned to say a DIV, or other elements like SPAN, TABLE ect?

machour:
@jakejackson1 this currently only works for `A`. I can look into adding the support for all other tags.

I'm not sure how to write the test for this change.. could you give me some advice?

finwe:
Closing this as stale and missing tests. Will gladly reopen after updated according to current contribuiton guidelines.

PR #661, opened 2018-02-26
refactor: Css manager cleanup

machour:
Usual stuff:

- removed unnecessary double quotes
- used short syntaxes / type casting
- strict strings comparison
- remove initialized variables that are immediately overwritten

`+`

- got rid of a bunch of unneeded regex (str(i)pos was enough)
- avoided count() in loops


finwe:
Merged in `development`, thanks!

PR #662, opened 2018-02-26
Clean DirectWrite

machour:
Usual stuff:
 * Removed unused variables
 * Type-safe comparisons where possible
 * single quotes instead of double
 * !empty($foo) instead of isset($foo) && $foo


finwe:
Merged in `development`, thanks!

PR #692, opened 2018-03-20
Avoid a Notice in BlockTag

sparkweb:
Currently getting Notice: Undefined offset: 1 in /mpdf/mpdf/src/Tag/BlockTag.php on line 174 in v 7.0.3 when I have a plain unstyled list inside a table. Doesn't always happen but in my tests it looks like it happens if I've got two items in the list. $this->mpdf->listtype comes in empty, thus causing the notice.

### DIFF
--- a/src/Tag/BlockTag.php
+++ b/src/Tag/BlockTag.php
@@ -173,7 +173,8 @@ abstract class BlockTag extends Tag
 				$decToAlpha = new DecToAlpha();
 				$decToRoman = new DecToRoman();
 
-				switch (isset($this->mpdf->listtype[$this->mpdf->listlvl]) ? $this->mpdf->listtype[$this->mpdf->listlvl] : "") {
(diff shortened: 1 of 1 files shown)


finwe:
Thanks for the PR. Can you add a unit test with HTML snippet failing before and passing after your change? See `tests/Issues` directory in the repo.

Also, can you move the changed statement outside the switch call for better readability? Use single quotes `'` for the empty string.

sparkweb:
I made the change as requested in the file itself. I was going to do this for the Unit Test, but I'm not quite sure how to add that to this PR:

```<?php

namespace Issues;

class Issue692Test extends \Mpdf\BaseMpdfTest
{

  public function testMissingMiwNotice()
	{
		$this->mpdf->WriteHTML('<ul><li>List 1</li><li>List 2</li></ul>');
		$output = $this->mpdf->output('', 'S');
		$this->assertStringStartsWith('%PDF-', $output);
	}
  
}
```

finwe:
This test passes even without your code change. Please provide HTML code sample failing without the change for the test.

finwe:
Closing as stale. Feel free to reopen with requested changes.

PR #711, opened 2018-04-03
Add correct visibilty

ColonelMoutarde:


Klap-in:
Please see #502 and #542, why these visibility is not yet updated. It's intentionally.

ColonelMoutarde:
"If you declare a property using var instead of one of public, protected, or private, then PHP 5 will treat the property as if it had been declared as public. "

from phpdoc http://php.net/manual/en/language.oop5.properties.php

finwe:
Thanks for the PRs, I am however going to point you again [to the previous discussion](https://github.com/mpdf/mpdf/pull/542#issuecomment-345658243) and close this.

There is nothing wrong on marking everything as public, I would just like it to be done properly right away rather than quick-patching this by marking everything public (unless of course future [language changes](https://wiki.php.net/rfc/var_deprecation) create a need for this).

Issue #724, opened 2018-04-20
Table header repeat bug 7.0.3 - Undefined index: y


sreenadh:
### I found this bug
Undefined index: y

Happens when a table has <thead> and the table does not fit on a page.

### This is mPDF and PHP version and environment (server/fpm/cli etc) I am using
mPdf version: 7.0.3
PHP 7.0.24

This has been working fine for 6.0.

#682 #516

finwe:
Fix merged, closing.

PR #725, opened 2018-04-20
Fix bug #724

sreenadh:


jakejackson1:
Thanks for your contribution. Mpdf still supports PHP5.6 so we are unable to use the null coalesce at this time. Also, all PRs require appropriate unit tests. Please see https://github.com/mpdf/mpdf/tree/development/tests/Issues for examples.

sreenadh:
@jakejackson1 Done 👍

finwe:
Why introducing a new static method instead of just using get with null as the third parameter?

finwe:
Thanks!

Issue #268, opened 2016-11-18
OverWrite not replacing text


mhoogstraten:
### I have this problem/would like to have this functionality
I want to replace text inside a PDF document (the basics), it now shows the original PDF without replacements.

### These are mPDF and PHP versions I am using
PHP 7.0.2 and mPDF 6.1.0

### This is a PHP code snippet I use

```
<?php
//Replace these to your file locations
$inputFileName = \Yii::getAlias('@app/tmp/pdf/tmp/imation.pdf');
$outputFileName = \Yii::getAlias('@app/tmp/pdf/tmp/imation2.pdf');

//Replace contents
$mPDF = new \mPDF();
$mPDF->SetImportUse();
$mPDF->percentSubset = 0;
$mPDF->OverWrite( $inputFileName, 'naam', 'Martijn', 'F', $outputFileName ) ;

```

[imation.pdf](https://github.com/mpdf/mpdf/files/600432/imation.pdf)

jiyanibharat303:
Please help I have checked MPDF 5.7, MPDF 6.0 and this for MPDF 7.0 and the overwrite is not working in any of them -
require_once __DIR__ . '/vendor/autoload.php';
$mpdf = new \Mpdf\Mpdf();
$mpdf->SetImportUse();
// forces no subsetting - otherwise the inserted characters may not be contained in a subset font
$mpdf->percentSubset = 0;
$mpdf->OverWrite('mpdf.pdf', 'Name', 'BHARAT', 'I', 'mpdf.pdf' ) ;

finwe:
Closing as duplicate of (better researched) #626 - uploaded example PDF has CR line endings which OverWrite has trouble recognizing as well

finwe:
Duplicate of #626

PR #745, opened 2018-05-11
fix overwrite

lalop:
May fix https://github.com/mpdf/mpdf/issues/268

Some texts can't be found with the current regular expression. I updated it to handle all the obj/endobj.

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -29373,55 +29373,48 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 		}
 
 		$changes = [];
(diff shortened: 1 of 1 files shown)


finwe:
Thanks for your PR! Can you accompany this with a unit test covering this behaviour? Also, fix travis errors and maintain code style.

lalop:
Ok for the code style, I could I test this kind of behaviour ?

finwe:
See tests in `tests/Issues` - test with a code/PDF sample which fails without this change. Upload a small template pdf to `tests/data/pdfs`.

PR #747, opened 2018-05-14
Overwrite pr 2

lalop:
I rewrite the pr and add a test.
If the replacement seems to work it can broke the rendered pdf.
For exemple the unit test try to replace `MAIN HEADING` per `replacement` in the test pdf
[2-Page-PDF_1_4.pdf](https://github.com/mpdf/mpdf/files/2000982/2-Page-PDF_1_4.pdf).
This result in this new pdf [test.pdf](https://github.com/mpdf/mpdf/files/2000984/test.pdf) where we can see that the text is correctly replaced but it appears to be crushed. Unfortunately I don't know too much pdf format to understand what happens here


lalop:
The new test is broken on a line I didn't change. I don't understand why this is broken, it works locally.
Is that possible the pdf isn't loaded on travis ?

PR #790, opened 2018-06-26
Issue788Test

TrungPhuNA:
This is the code I wrote and it was corrupted. I think it's because of css.


finwe:
Cannot use this, the test has to contain pure HTML, mPDF itself cannot use view() helper and the template itself is also missing.

Issue #578, opened 2017-12-04
Feature: setasign/fpdi: 2.0 support


polukes:
Please add support to setasign/fpdi: 2.0 version. 
For example I cannot use mPDF with newest version of iio/libmergepdf.

finwe:
Related to https://github.com/mpdf/mpdf/issues/480

mpochron:
+1

hella-schu:
Yes please!

do-while:
+1

azbatuk:
+1

thbl:
+1

JanSlabon:
I'm on it: https://github.com/JanSlabon/mpdf/tree/FPDI-2-Update

JanSlabon:
May be the case but please do not use this in production! This was my first commit to the upgrade and it was tested only with the tests suite of mPDF while this part is only covered sporadic ;-)

Y0lk:
Looking forward to this. Any update?

commonpike:
I cant use FPDF+PDFI when using mPDF because of these dependencies..  dropping mPDF for now I'm afraid.

JanSlabon:
The pull request is [there](https://github.com/mpdf/mpdf/pull/899). Feedback and/or discussion is welcome.

commonpike:
Yes, and that's great. But you said 'don't use this in production' .. 15 files changed, so I'm not sure what the status is..

donalturrentine:
I don't use ImportPage in my application and I do extensive testing for the use cases... I'm happily using my fork in production via composer.

JanSlabon:
@commonpike From my point of view the final pull request is stable. @donalturrentine just commented directly after I did my first commit, which was not tested nor was it working as expected! If he still uses the fork after the first commit, it is completely broken and I urgently suggest not to use it. The resulting PDFs were faulty - still some viewers were able to display them. But under the hood they were simply broken. Upgrade to the current state of the pull request and everything should be fine. Feedback is welcome!

commonpike:
Thanks for the clarification, and the PR.

I'm trying to use it, but I can't figure it out :-/ I am stuck to using composer to get the branch of your PR into my app, which works, but somehow the psr-4 autoloaders  don't seem to pick it up. `new \Mpdf\Mpdf();` says class not found and so does `new Mpdf();` ..

.. still just hoping this is merged anytime soon, that would be great.

davidjosephhayes:
[...]

PR #796, opened 2018-06-28
Update composer.json

alef-carvalho:
Update composer.json to resolve composer installation conflict with other packages


jakejackson1:
Updating to FPDI v2 is quite a bit more involved that just changing the composer package, as `Mpdf.php` extends some of its functionality and these parts will need to be rewritten.

finwe:
See https://github.com/mpdf/mpdf/issues/578

Issue #205, opened 2016-07-29
Background image is blurry in Firefox


tifischer:
### I have this problem/would like to have this functionality

When the generated PDF is viewed in the FireFox browser, the background image is greatly distorted as a blur.
### These are mPDF and PHP versions I am using

mPDF 6.1.2 and PHP 5.5.31
### This is a PHP code snippet I use

```
<?php

$tableData = "<div id=\"cardLogo\" style=\"width:3.325in;height:2.125in;margin:0.125in; background-image: url(logo.jpg); background-repeat:no-repeat; background-size: 2in 3.25in; background-position: center;  \"> 
    <div style=\"float:left; margin-left:.4in; height:1.125in; width: .875in; padding-top:.8in; \"><img src=\"logo.jpg\" /></div>
    <div style=\"float:left; margin-left:.05in; width:.7in; padding-top:.75in;\">
    <table>
            <tr>
                <td></td></tr></table></div></div>";
```

jakejackson1:
If the background image looks fine in Adobe Reader the issue is with Mozilla's pdf.js library and you should [open an issue over on their GitHub repo](https://github.com/mozilla/pdf.js/).

RhRU:
If anybody has this issue in FF one trick is to remove the no-repeat from the background

honsa:
It happens in chrome too, background-image get blured if background-repeat is set to repeat.

mPDF 7.0.0
PHP 5.6.18

jBesic:
The issue is caused by the value for XStep and YStep and how the pdf.js handle these huge values. I have tested this within my project by changing the following and it works well.

`$this->_out(sprintf('/YStep %d', 99999));` => `$this->_out(sprintf('/YStep %d', 512));`
`$this->_out(sprintf('/XStep %d', 99999));` => `$this->_out(sprintf('/XStep %d', 512));`

RhRU:
@jBesic (pozdrav) Create a push request?

Klap-in:
Please add an unit test reproducing this case if you open a pull request. For examples you could have a look at https://github.com/mpdf/mpdf/tree/development/tests/Issues Thanks!

finwe:
@Klap-in please when you do, comment directly to the proposed PR, do not encourage issue hijacking.

Klap-in:
Sorry! Valid point!

LubomirKurpel:
> If anybody has this issue in FF one trick is to remove the no-repeat from the background

[...]

PR #807, opened 2018-07-10
#205 Change desired horizontal and vertical spacing between pattern c…

jBesic:
Decrease the value for the XStep and the YStep properties from `99999` to `512`.


Klap-in:
Can you complete the pull request with an unit test? For examples of unit tests you can have a look at https://github.com/mpdf/mpdf/tree/development/tests/Issues
Thanks in advance!

PR #833, opened 2018-08-10
Fix for issue #556 - background-size: cover CSS size computation

### LINKED ISSUE #556: 'background-position: center' & 'background-size: cover' stretching image
i copied this issue from mpdf.github.io (i dont know what the difference between these 2 githubs is? but i posted it there first)

the css attributes 'background-position: center' & 'background-size: cover' on an element with a background-image only work, as long as the background image gets cropped vertically, but if it would get cropped horizontally, the image will instead be stretched and repeated.

[this is how it looks like & should look like in Html, and how its rendered in the pdf](https://imgur.com/a/QZcL1)

Does anyone know a workaround for this? the elements with the background-image also have width & height = 100% and im assuming it could have something to do with those percent sizes, but i need them to be 100%.

scm6079:
Computation of the image size for the background-size: cover mode always assumed a fixed aspect ratio between image and viewport (e.g. never a landscape image in a portrait viewport).  This code now looks at both image and viewport aspect ratio to compute image size.

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -2387,7 +2387,7 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 								// and supporting both tall or wide images in relation to the area
 								$aspectImage = ($ih!=0)?($iw / $ih):1;
 								$aspectViewport = ($pb['bpa']['h']!=0)?($pb['bpa']['w'] / $pb['bpa']['h']):1;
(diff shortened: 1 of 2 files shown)


krzyc:
This fixes background-size: cover bug for me. What is needed to merge this PR?

finwe:
Given how old it is, probably rebasing it on current development branch and fixing potential tests and CS issues.

PR #841, opened 2018-08-16
Allow the Fonts to be loaded via Packages

jakejackson1:
This is the first attempt at decoupling the fonts from Mpdf. The code / fonts pulled out from this PR can be found here: https://github.com/jakejackson1/mpdf-unicode. In the separate package is the fonts, a font configuration file, and a language to font configuration file.

Before I spent too much time on it, I wanted to open this up to discussion about if this is the right direction to take. I envision we split up the current fonts package into a number of smaller modules that cater to the different script types: unicode (name chosen because of the vast array of characters Dejavu and Freesans supports), CJK, Arabic, Egyptian ect. We can then add each as a suggested composer package. The benefit to this approach (besides reducing the size of core) is it'll allow users interested in each language to maintain fonts suitable for their language (Bengali comes to mind) without effecting the core. 

For those that would like to test this out, the only change required is to use the new `fontRegistry` configuration key when you initialise mPDF:

```
$mpdf = new Mpdf\Mpdf([
	'fontRegistry' => new \Mpdf\Fonts\FontRegistry([
		new \Mpdf\Fonts\Unicode\Config\FontVariables(),
	]),
]);
```
I thought about loading known packages automatically, but for now opted for manual initialisation. We can certainly explore the automated option if enough people like the idea. 

From a technical perspective, this PR does not negatively effect any of the existing configuration keys to do with fonts and it all works together seamlessly (or that's the idea anyway - no tests yet).


src/Mpdf.php
```diff
@@ -1587,12 +1593,51 @@ private function initConstructorParams(array $config)
...
+					'backupSubsFont' => $package->getBackupSubsFont(),
+				] + $package->getFontFamilySubstitution();
+
+				foreach( $packageFontData as $id => $val ) {
+					if (isset($config[$id])) {
+						$config[$id] = array_merge($config[$id], $val );
+					} else {
+						$config[$id] = array_merge($val, $defaults[$id]);
+					}
+				}
```
machour: `array_merge()` should be avoided in loops (and we're in two loops)

would it be possible to gather data in the loop instead, and perform a single array_merge at the end ? 🤔

src/Language/LanguageToFontRegistry.php
```diff
@@ -0,0 +1,111 @@
...
+	{
+		$font         = '';
+		$coreSuitable = false;
+
+		foreach ($this->getAll() as $class) {
+			$results = $class->getLanguageOptions($llcc, $adobeCJK);
+
+			if (is_array($results)) {
+				$font         = strlen($results[0]) > 0 ? $results[0] : $font;
+				$coreSuitable = $results[1] === true ? true : false;
```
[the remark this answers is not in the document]
machour: From what I see, `getLanguageOptions` returns an array, why are you testing for type below?

If the return is mixed, could we change it to be always an array ?
If it's complicated, can we first convert $results to an array, and then always apply the same treatment to make the code more readable ?

i.e.

```
if (!is_array($results)) {
   $results = [$results];
}

$font = strlen($results[0]) > 0 ? $results[0] : $font;
$coreSuitable = $results[1] === true;
```

src/Language/LanguageToFontRegistry.php
```diff
@@ -0,0 +1,111 @@
...
+	 *
+	 * @param $classes
+	 *
+	 * @since 8.0
+	 */
+	public function __construct(array $classes)
+	{
+		array_walk($classes, function ($class, $key) {
+			$this->add($key, $class);
+		});
```
machour: why not use a simple `foreach` loop?

src/Fonts/FontRegistry.php
```diff
@@ -0,0 +1,86 @@
...
+	 * @param FontRegistrationInterface[]|FontRegistrationInterface $classes
+	 * @since 8.0
+	 */
+	public function __construct($classes)
+	{
+		$classes = is_array($classes) ? $classes : [$classes];
+
+		array_map(function ($class) {
+			$this->add($class);
+		}, $classes);
```
machour: why not use a simple `foreach` loop ?

finwe:
I generally like this idea, just a few remarks for now:

- I'd very much like to keep at least Dejavu along with the main package - UTF support out of the box is one of the main mPDF selling points. Sure, these fonts can be split, but then the package should be a root dependency of the main package
- The configuration is hands down clean and extendable, yet it seems a bit too complex for a newcomer - could we come up with a simpler way of registering a new font than to create multiple classes and linking them together?
- How about some "font autodiscovery"? So that only requiring a composer package is needed. That would simplify at least including of already existing font packages.

machour:
@finwe I'm not really sure about your first bullet point, as we're talking about 19M (it just took me 26 seconds to clone Jake's repo). If the end user doesn't need them, then it's a pure waste.

Could we instead update the installation guide with heads-up, and use Composer's [suggest](https://getcomposer.org/doc/04-schema.md#suggest) feature to hint the user about available packages instead ?

finwe:
Will give it some more thought.

src/Language/LanguageToFontRegistry.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @param string                  $name
+	 * @param LanguageToFontInterface $class
+	 */
+	public function add($name, LanguageToFontInterface $class)
+	{
+		$this->register[$name] = $class;
+	}
+
+	/**
+	 * Remove a anguageToFont Package by Name
```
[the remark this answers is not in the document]
jakejackson1: Thanks.

src/Language/LanguageToFontRegistry.php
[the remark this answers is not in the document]
jakejackson1: For the majority of the font packages that will be created, the `coreSuitable` property won't be required. For simplicity for package maintainers, I opted to accept both an array and string. However, I've reworked this function based on your feedback.

src/Language/LanguageToFontRegistry.php
jakejackson1: I must have written this right after working in JavaScript where "functional" programming is all the rage.

src/Mpdf.php
jakejackson1: Not sure if that'll be possible. Would love it if you could take a closer look and brainstorm some ideas.

jakejackson1:
I've rebased this PR on the latest development branch and tidied up a couple things based on @machour feedback (thanks).

@finwe 

* I'm with @machour on this, let's strip all the fonts out of core and allow users to pick and choose what font packages they want (or just use the Core mPDF fonts)

* The auto-discovery [feature in Aura](http://auraphp.com/blog/2014/04/07/two-stage-config/) looks better than [the Laravel version](https://divinglaravel.com/laravels-package-auto-discovery). I'll have a tinker when I can and look at implementing it.

acasademont:
Hi! Is this still relevant? We're using Docker and right now mpdf is by far the fattest package due to the fonts directory. Composer's `suggest` would be the best way to proceed imho!

finwe:
Yes, it is relevant. Yes, composer suggest for the most common font packages will be in place after the mechanism is finalized and fool-proofed.

finwe:
@jakejackson1 I believe you wanted to reference #2161

PR #847, opened 2018-08-23
Adds support for text-align: start and end

josephzidell:
We need to produce documents in LTR as well as RTL using the same markup, and this PR adds support for: `text-align: start` and `text-align: end`


josephzidell:
Anything else I can do to get this merged?

Klap-in:
It would be good to add unit tests for this feature as well. Please see for examples https://github.com/mpdf/mpdf/tree/development/tests.

PR #854, opened 2018-08-28
clear buffer

TretyakovArtem:
Trying to fix memory leak when using in long-running task.

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -9385,6 +9385,8 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 
 				fwrite($f, $this->buffer, strlen($this->buffer));
 				fclose($f);
+				$this->cache->clearOld();
+				$this->buffer = null;
 
 				break;
 


finwe:
Please, elaborate, test (with PHPUnit) and verify that this has any impact.

Wild long shots like these have no place upstream, sorry.

Feel free to reopen the PR with additional tests and reasoning.

seme1:
I've had a nasty memory leak with the output() function that was only fixed after adding these two lines:

				$this->cache->clearOld();
				$this->buffer = null;

This should definitely be merged.

TretyakovArtem:
> I've had a nasty memory leak with the output() function that was only fixed after adding these two lines:
> 
> ```
> 			$this->cache->clearOld();
> 			$this->buffer = null;
> ```
> 
> Well, it seems that the problem randomly goes away and returns back. The above code did not help :(

Do you run your app in docker container? We have found problem, when check memory usage inside docker container. Memory does not leak. Our monitoring system was wrong.

seme1:
No, I am not using Docker.

I simply have the below function that is called repeatedly inside a for loop. There is a memory leak caused by the usage of mpdf even though it should be destructed by the end of the function. I'm unable to run this function on past the 170 files.
```
function writeToFile($stylesheet,$InvContent,$output){
$mpdf = new \Mpdf\Mpdf([
      'fontDir' => array_merge($fontDirs, [ROOT. 'css/DubaiFont/orig',]),
      'fontdata' => $fontData + [
          'frutiger' => [
              'R' => 'Tajawal-Regular.ttf',
              'B'=>'Tajawal-Bold.ttf',
              'useOTL' => 0xFF,
              'useKashida' => 75,
          ]
      ],
      'default_font' => 'frutiger'
  ]);

    $mpdf->autoScriptToLang = true;
    $mpdf->baseScript = 1;

    $mpdf->autoArabic = true;
    $mpdf->autoLangToFont = true;

   $mpdf->WriteHTML($stylesheet, \Mpdf\HTMLParserMode::HEADER_CSS);
   $mpdf->WriteHTML($InvContent, \Mpdf\HTMLParserMode::HTML_BODY);
   $mpdf->Output($output);
   unset($mpdf);
   unset($stylesheet);
   gc_collect_cycles();
}
```

Issue #871, opened 2018-09-28
setasign/fpdi: 1.6.* not compatible with PHP 7.2


azbatuk:
One of mPDF's dependencies, "setasign/fpdi": "1.6.*", is using a PHP function, "each()", that has been deprecated from PHP 7.2. The latest version of fpdi is 2.1.0. Would it be possible to use the latest "setasign/fpdi" by simply replacing the version number in the composer.json with "2.1" ? If not, what other solution may we have ?

I am using mPDF 7.1.5 and PHP 7.2 running on Ubuntu 18.04.

Thanks

finwe:
There already is a feature request for fpdi 2.x support. AFAIK major versions of fpdi are not compatible, so direct replacement is not possible.

Can you please share a simple code example demonstrating this behaviour on PHP 7.2?

JanSlabon:
2.x support is still on our list and this issue gives it more priority. I guess we'll do this within the next 2-3 weeks (due to holidays over here).

do-while:
+1

finwe:
Closing in favor of #578

PR #878, opened 2018-10-04
Allow use of setasign/fpdi greater than 1.6

donalturrentine:
composer.json is set to require setasign/fpdi 1.6.*, this would allow it to use a minimum of 1.6 and allow newer. Needed for compatibility with PHP7

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
 		"ext-mbstring": "*",
 
 		"psr/log": "^1.0",
-		"setasign/fpdi": "1.6.*",
+		"setasign/fpdi": "^1.6",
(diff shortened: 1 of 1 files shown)


finwe:
There is no point in this change. 1.x branch of FPDI ends with 1.6.2 version, upgrade to 2.x branch is on the way https://github.com/mpdf/mpdf/issues/578

donalturrentine:
> There is no point in this change. 1.x branch of FPDI ends with 1.6.2 version, upgrade to 2.x branch is on the way #578

Thanks, that will work! What is the timeline for 2.x?

finwe:
There is none. Hope setasign will have time to look into this https://github.com/mpdf/mpdf/issues/871#issuecomment-425846450

Issue #270, opened 2016-11-23
Time is changed when PDF is generating


1manfactory:
### I have this problem/would like to have this functionality
I am logging all my generating processes to keep track of errors. I came across a jump back in time when mpdf is called.

### These are mPDF and PHP versions I am using
mPDF 6.1
PHP 5.4.45

### This is a PHP code snippet I use

```
$started=date("H:i:s");
print $started."<br>\n";

require_once '../../mpdf61/mpdf.php';

$mpdf = new Mpdf();
$mpdf->WriteHTML('<p style="font-size:4em;">Started: '.$started.'</p>');
$mpdf->WriteHTML('<p style="font-size:4em;">Now: '.date("H:i:s").'</p>');
$mpdf->Output('./time.pdf');

print date("H:i:s")."<br>\n";
?>

```
This is the output on my screen:
```
19:40:30
18:40:30
```

It's the same in the attached file.
[time.pdf](https://github.com/mpdf/mpdf/files/609724/time.pdf)

And to make it worse, the process keeps it's now wrong time so even when I am finished with generating PDF files I am still on the wrong time.

1manfactory:
Ok. I had to fix it with setting the time zone back again immediately after including mpdf. But it's still strange.

> require_once '../../mpdf61/mpdf.php';
> date_default_timezone_set('Europe/Berlin');

finwe:
Yes, mPDF up to 6.1 changes the timezone to prevent warnings about the timezone not being set. This is now removed in development.

Issue #137, opened 2016-05-25
Directive for date.timezone


duskohu:
Hi, why do you have this directive?
```
if(function_exists("date_default_timezone_set")) {
	if (ini_get("date.timezone")=="") { date_default_timezone_set("Europe/London"); }
}
```
For example, when I have on server:
```
var_dump(date_default_timezone_get()); >> Europe/Prague
var_dump(ini_get("date.timezone")); >> false
```
This directive change my timezone.

finwe:
see https://github.com/mpdf/mpdf/issues/27, the code will be removed in future major version.

Issue #27, opened 2015-06-12
Mpdf change timezone even if timezone is set


kravcik:
I made some step-by-step test, where is problem with timezone on production server. 

1.
Server setting timezone "Europe\Prague"
date_default_timezone_get(); //Europe/Prague

2.
Framework (Nette) setting timezone "Europe\Prague"
date_default_timezone_get(); //Europe/Prague

3.
Loading mPdf by autoload
date_default_timezone_get(); //Europe/London

4.
Comment lanes:
//if(function_exists("date_default_timezone_set")) {
	//if (ini_get("date.timezone")=="") { date_default_timezone_set("Europe/London"); }
//}
date_default_timezone_get(); //Europe/Prague

rzds:
I have the same problem with mPDF.
Can you please explain why you need to set a default timezone inside the library?

finwe:
I don't know the reason that the original author of the library had. This really is a bug (or better an undesirable behaviour) and entire timezone alteration by the library will be removed in the future versions.

culunvb:
hi all, i have the same problem, create a new variable with date("Y-m-d") before i defined mpdf class, it work's fine.

maybe this help u all
thx

PR #892, opened 2018-10-23
Remove timezone override, to avoid conflict when mpdf is used with other libraries

enriquelacoma:
As said on issues:

- https://github.com/mpdf/mpdf/issues/27
- https://github.com/mpdf/mpdf/issues/270
- https://github.com/mpdf/mpdf/issues/137

The file mpdf.php forces the timezone, we have found that this causes problems when we use this with other libraries, for example on this drupal module https://www.drupal.org/project/print overrides the timezone of the site.

### DIFF
--- a/mpdf.php
+++ b/mpdf.php
@@ -90,12 +90,6 @@ $errorlevel = error_reporting($errorlevel & ~E_NOTICE);
 
 //error_reporting(E_ALL);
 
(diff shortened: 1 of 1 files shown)


finwe:
6.x branch is frozen and changes are no longer accepted. Upgrade to version 7 where this is already solved.

PR #897, opened 2018-10-29
Allow the non-active form radio field color to be set with CSS

jakejackson1:
The non-active checkbox field is more difficult to control via CSS due to it making use of two colours: a white square and a black cross. Because of this, I've only focused on the radio field.


jakejackson1:
Unit test added.

Issue #608, opened 2018-01-11
mPDF update FPDI Parser to Commercial Version


fsmalaquias:
> Please use https://stackoverflow.com/questions/tagged/mpdf for all your general questions or troubleshooting!
> For contributing here, please see the guideline: https://github.com/mpdf/mpdf/blob/development/.github/CONTRIBUTING.md

Hi,

I'm trying to user FPDI Parser Commercial Version that uses FPDI 2.0.*

I updated mpdf composer file to user FPDI Parser 2.0 but I think that mPDF 7.0.3 is not prepared to use it.

How can I help?!

mPDF needs to have a version with FPDI Parser Commercial

This is the manual for upgrade
https://manuals.setasign.com/fpdi-manual/v2/migrating/

finwe:
See also https://github.com/mpdf/mpdf/issues/480

JanSlabon:
You can use this again as of mPDF 8. @finwe I guess you can close this one :-)

Issue #480, opened 2017-08-31
importPage() parameter are missleading and fault-prone


JanSlabon:
We're currently preparing a new version of FPDI for the public (https://github.com/Setasign/FPDI/releases/tag/v2.0.0-beta2).

In the near future we would like to make mPDF compatible with this version of FPDI but actually I'm of two minds with the parameter from importPage():

    function ImportPage($pageno = 1, $crop_x = null, $crop_y = null, $crop_w = 0, $crop_h = 0, $boxName = '/CropBox')

I understand what Ian would like to accomplish with the $crop_* parameters but he didn't thought about boundaries which do not have their origin on [0,0] or a page that is rotated through its Rotate entry in the page dictionary. An example document is here: [[-1000 -1000 -500 -500].pdf](https://github.com/mpdf/mpdf/files/1268392/-1000.-1000.-500.-500.pdf)

The $crop_* parameters are replacement values for the page boundary values of the imported page and do not crop the imported page directly. To accomplish this it would be required to calculate the relative values to the imported page boundary. Also the rotation of the imported page shall be taken into account... after all this is much effort for a feature that can be done much easier in the graphic state of a page, before the imported page is used (such as by _setClippingPath()). Over all we never got a request for such feature on the offical FPDI version for more than a decade.

Would it be ok, to remove the $crop_* parameter in an upcoming pull request or will this be a no-go?

finwe:
How about introducing a new method and marking ImportPage as deprecated until v 8? Shame to lose the method name, though.

JanSlabon:
A new method name would be somewhat odd.

Isn't a new major version (7) made for such changes? I mean by removing the parameters we remove a faulty behavior. Overall I guess that this "feature" was not used very often.

After all it would be possible to keep the behavior but as it is wrong it needs some overcoming at our end :-)

finwe:
Sure, we _can_ introduce breaking changes in major version per semver "spec", however, I would like the new method to be error-prone and to warn when being used the old way so that people upgrading know when they missed an occurence in their code. 

How would you propose to do that? Throwing an error/exception when there is a second parameter that is not a string? when there are more than two parameters? All these solutions  are not 100% foolproof. I'd rather not introduce another fault-prone state by (insufficiently) fixing another.

JanSlabon:
Isn't this enough:

    if ($box === null || is_numeric($box)) {
        throw new \InvalidArgumentException('...');
    }

?

[...]

PR #899, opened 2018-10-30
Update to FPDI 2

### LINKED ISSUE #578: Feature: setasign/fpdi: 2.0 support

JanSlabon:
# BC breaks

- Removed SetImportUse() method.
- Signature of importPage() changed to default signature of FPDI (related to #480). Currently no alternative is implemented.

This fixes #578, #608 

Please review and comment. I am out of the office until 5th November.


src/FpdiTrait.php
```diff
@@ -0,0 +1,263 @@
...
+	 * A counter for template ids.
+	 *
+	 * @var int
+	 */
+	protected $templateId = 0;
+
+	protected function setPageFormat($format, $orientation)
+	{
+		// in mPDF this needs to be "P" (why ever)
+		$orientation = 'P';
```
JanSlabon: This is a very strange behavior of mPDF and I didn't found any other reasonable way to "fix" this.

src/Writer/ResourceWriter.php
```diff
@@ -174,38 +164,14 @@ public function writeResources() // _putresources
...
-								foreach ($v[1] as $k2 => $v2) {
-									$this->writer->write($k2 . ' ', false);
-									$this->mpdf->pdf_write_value($v2);
-								}
-							}
-						}
-					}
-				}
-			}
-
```
JanSlabon: We never received any report for FPDI regarding such behavior/issue. Maybe it was an issue in a viewer application but from the technical view this is simply not required.

JanSlabon:
Would it make sense to re-introduce the fault signature of importPage() mentioned in #480 to get this into the 7.x version? I mean this could be possible but at least I have no idea to write a test for it as it's simply wrong :-) 

Is there a timing for version 8?

JanSlabon:
Exactly. FPDI v1 cannot be used with this pull request.

rostockahoi:
Hi, I have trouble processing some PDFs with mPDF 7.1. I get an exception which guides me to https://www.setasign.com/products/fpdi-pdf-parser/details.

```
This document (/data/www/html/**/qo1qhe.pdf) probably uses a compression technique which is not supported by the free parser shipped with FPDI. (See https://www.setasign.com/fpdi-pdf-parser for more details)
```

Could this be solved by updating to FPDI 2 and using the commercial FPDI parser?

Thank you.

finwe:
@rostockahoi Please do not abuse this conversion for not-entirely-related general questions.

finwe:
@JanSlabon This looks OK. Can you also please look at examples repo? The old ImportPage signature is used and you'll certainly be faster in changing the calls to end in a same result as before.

JanSlabon:
Done: https://github.com/mpdf/mpdf-examples/pull/20

finwe:
Awesome, thanks.

PR #911, opened 2018-11-09
chore(WriteHTML): Make WriteHTML use const instead of magic numbers

tomtomau:
Hey! Thanks for this package, it works quite well.

I was pleased to see that `Destination::FILE` constants exist but was disappointed to see the same didn't exist for the magic numbers which defined the second argument to `WriteHTML()`

I specifically referred to the [external documentation](https://mpdf.github.io/reference/mpdf-functions/writehtml.html#parameters) while doing this as well:

* rename `$sub` to `$mode`, I don't know what sub was meant to mean but I don't see why it shouldn't be called `$mode`
* replace any checks that relied on the order of the constants (ie `$sub < 1) with their true meanings to aid readability
* enforced strict equality checks `===` and enforced that the mode should be one of the permitted modes
* I named the constants the best I could think of but more than happy to change that based on having a better understanding than me about what they do :D
* wrote tests to cover the enforcement of the modes

I'm happy to update docs if accepted too

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -9550,7 +9550,7 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 				$this->pageBackgrounds = [];
 
 				$this->writingHTMLheader = true;
-				$this->WriteHTML($html, 4); // parameter 4 saves output to $this->headerbuffer
+				$this->WriteHTML($html, HTMLParserMode::HTML_HEADER_BUFFER);
 				$this->writingHTMLheader = false;
 				$this->Reset();
 				$this->pageoutput[$n] = [];
@@ -9633,7 +9633,7 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 
 				$this->writingHTMLfooter = true;
 				$this->InFooter = true;
-				$this->WriteHTML($html, 4); // parameter 4 saves output to $this->headerbuffer
+				$this->WriteHTML($html, HTMLParserMode::HTML_HEADER_BUFFER);
 				$this->InFooter = false;
 				$this->Reset();
 				$this->pageoutput[$n] = [];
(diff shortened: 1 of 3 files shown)


src/Mpdf.php
```diff
@@ -12964,6 +12962,10 @@ function WriteHTML($html, $sub = 0, $init = true, $close = true)
 			}
 		}
 
+		if (in_array($mode, HTMLParserMode::getAllModes(), true) === false) {
+			throw new \Mpdf\MpdfException('WriteHTML() requires $mode to be one of the modes defined in HTMLParserMode');
```
finwe: Can we, for the time being, use a less-strict warning here? Eg throwing an exception only when the debug mode is on? This is IMHO somewhat a BC break, so I'd be reluctant to incorporate this in a minor release.

jakejackson1: It might also be a good idea to default `$mode` to `HTMLParserMode::DEFAULT_MODE` if the user passes in an unknown mode for backwards compatibility reasons.

tomtomau:
Hey @finwe and @jakejackson1, thanks for the speedy response!

I've updated the PR, I initially tried turning off the strict equality check in_array, but man does it lead to some [weird results](https://3v4l.org/mOcIj) - so I think it's still worth keeping.

I've made it only throw an exception if debug is enabled, otherwise use the default mode and added tests to cover this.

I've also rebased back ontop of development to resolve some merge conflicts.

After this is merged, I'm keen to create another PR for making it as strict as the pr originally outlined (always throwing exceptions) - is there a BC break release planned that could get rolled into?

Appreciate the work y'all doing to maintain and improve such a beast of a project, PHPStorm was having a meltdown when it opened Mpdf.php 🙃

finwe:
Thanks for this and sorry for the delay. It would be awesome if you were able to send a PR to the docs as well.

PR #915, opened 2018-11-15
chore(WriteHTML): Make WriteHTML mode checking strict [bc break]

tomtomau:
Per #911 - this pull request makes the mode checking more strict by always throwing exceptions, even when debug is disabled.

As discussed, this is potentially a bc issue for some people so this should target a BC break release, not sure if there's a more appropriate branch I can set as the base for this to be merged into.


finwe:
Merged to 8.0-RC1, thanks

PR #916, opened 2018-11-19
Remove unecessary ternary expressions

carusogabriel:


finwe:
Wouldn't bitwise return an integer, so at least a bool cast would be in order to preserve booleans?

carusogabriel:
@finwe You are right, this refactoring isn't right :+1:

PR #917, opened 2018-11-19
Use dedicated PHPUnit assertions

carusogabriel:
This will help us with better error messages when debugging errors.


finwe:
Cool, thanks!

PR #927, opened 2018-11-30
Use the img_dpi setting for all images

jakejackson1:
Any size conversion related to images should [use the img_dpi setting](https://mpdf.github.io/reference/mpdf-variables/img-dpi.html) instead of [dpi](https://mpdf.github.io/reference/mpdf-variables/dpi.html).

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -8700,8 +8700,8 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 			$info = $this->imageProcessor->getImage($this->noImageFile);
 			if ($info) {
 				$file = $this->noImageFile;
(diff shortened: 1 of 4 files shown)


finwe:
Isn't this a potential BC break? Wouldn't it make sense to remove img_dpi altogether in v8 for better clarity?

jakejackson1:
Technically, SVGs and text are vectors and don't need to bother with DPI since they can scale without any resolution issues. For example, I regularly will set the `img_dpi` to 300, but leave the `dpi` as is, so I don't have to rescale all my text at the same time.  

If we were to remove any option, it would make more sense to remove the `dpi` setting and use a constant 96 dpi for the current `dpi` setting. 

With that in mind, I've scaled back this PR to only apply to the `noImageFile` fallback image which is a jpg.

PR #934, opened 2018-12-05
explain purpose

Klap-in:
let see if we get more complete example? or is this too much?


finwe:
I doubt this would help. 80 % of issue reporters remove the template anyway.

PR #935, opened 2018-12-11
Convert Font Cache from PHP to JSON Files

jakejackson1:
This PR standardised the font cache. Instead of crudely generating PHP files with font data, it stores the information in JSON files. 

JSON cache files has a number of benefits:

1. The code to generate the files is much smaller. 
1. The data doesn't need to be escaped
1. It is easier to test and modify
1. More compatible on shared hosting with security that limit the dynamic creation of PHP files

**Backwards Compatibility**
This is fully backwards compatible. There is a one-time font cache refresh to generate the new JSON files, then it's business as usual. The font cache doesn't need flushing before hand.

I tested this PR against the Mpdf Example folder and found no visual regressions. 

**Benchmarks**
By forgoing the PHP cache, we lose OPCaching on these files, but I've kept the JSON cache performant none-the-less and there's little change in performance. 

Average (5 runs) for all [Mpdf Example Templates](https://github.com/mpdf/mpdf-examples/) using the CLI:
* PHP Cache: 147.839
* JSON Cache: 148.221

Difference: 0.26%

Average (5 runs) for [Example 61](https://github.com/mpdf/mpdf-examples/blob/development/example61_new_mPDF_v6-0_features.php) in the browser with OPCache Enabled:
* PHP Cache: 21.036661
* JSON Cache: 21.271170

Difference: 1.11%

**Bugs**
A few PHP notices were fixed during this revision. See `src/TTFontFile.php` and `utils/font_dump.php`. Note: I noticed all the other scripts in `utils` no longer work and need to be refined (or removed).

### DIFF
--- a/src/Fonts/FontCache.php
+++ b/src/Fonts/FontCache.php
@@ -7,6 +7,8 @@ use Mpdf\Cache;
 class FontCache
 {
 
+	private $memoryCache = [];
+
 	private $cache;
 
 	public function __construct(Cache $cache)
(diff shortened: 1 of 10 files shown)


finwe:
Awesome. Thanks a lot.

PR #940, opened 2018-12-14
fix Warning: A non-numeric value encountered in svg

luizsene:
added validation before the division operation

when sending an image of the higchart for pdf export the rgba conversion (192,192,192,0,0.001) of the stroke attribute in the function> colorConverter-> convert returns "." causing division error in PHP 7.2

### DIFF
--- a/src/Image/Svg.php
+++ b/src/Image/Svg.php
@@ -1369,10 +1369,10 @@ class Svg
 		else if (strtolower($critere_style['fill']) == 'currentcolor' && $element != 'line') {
 			$col = $this->colorConverter->convert($critere_style['color'], $this->mpdf->PDFAXwarnings);
 			if ($col) {
-				if ($col{0} == 5) {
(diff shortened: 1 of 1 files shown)


finwe:
Thanks for the PR! Can you please add a unit test with code failing without this change? See tests/Issues for reference.

RafaelGSS:
Hello everyone, I need this fix please...

finwe:
Hello everyone, I need a code sample please...

RafaelGSS:
But, `\Mpdf\BaseMpdfTest` have an assert for Warning? The code sample has in my issue #990

finwe:
Now without trolling: if you need a fix quickly, prepare a new PR with a test (failing prior to the code change and passing after). No need to assert a warning. See other tests in tests/Issues directory.

RafaelGSS:
Ok... creating. Thanks.

finwe:
Resolved by #996

PR #958, opened 2019-01-22
fix unset ws bug

nisests:


finwe:
Please add a test case with code affected by this change. see test/Issues directory for reference.

Also please adhere to the coding standard (use tabs, not spaces).

mshYat:
@finwe Dublicate of https://github.com/mpdf/mpdf/pull/1140.

PR #967, opened 2019-02-13
Change the Wildcard Version Range (.*) to Caret Version Range (^) in the package setasign/fpdi

lfluvisotto:
The package setasign/fpdi as Wildcard Version Range (.*) on 1.6.* is blocking other packages like tig-nl/postnl-magento2, so I changed to Caret Version Range (^) ^1.6

https://github.com/tig-nl/postnl-magento2/blob/v1.7.3/composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
 		"ext-mbstring": "*",
 
 		"psr/log": "^1.0",
-		"setasign/fpdi": "1.6.*",
+		"setasign/fpdi": "^1.6",
 		"paragonie/random_compat": "^1.4|^2.0|9.99.99",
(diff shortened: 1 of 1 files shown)


finwe:
Blocking how? 1.6 is the final minor version in 1.x line for setasign/fpdi so there is no actual difference between ^1.6 and 1.6.*

finwe:
Also, there is no public package `tig-nl/postnl-magento2` so I see no reason to change to suit every private package needs.

Feel free to reopen with more reasoning.

lfluvisotto:
@finwe sorry, I was in a rush and I forgot some words in the description what I meant was "blocking the update" other packages.

I'll open other pull request to discuss the case.

PR #968, opened 2019-02-14
Update package setasign/fpdi to ^2.1

lfluvisotto:
@finwe 

I'm working in a project which uses the packages "mpdf/mpdf" and "tig-nl/postnl-magento2"

I was trying to update the package "tig-nl/postnl-magento2" to the version 1.7.3 but I couldn't because this package is using the package "setasign/fpdi-fpdf": "2.1", which uses the "setasign/fpdi": "^2.1"

I'm not being able to do this update because "mpdf/mpdf" uses "setasign/fpdi": "1.6.*"

Let me know if I was clear and it makes sense to you.

I don't know if it'll be possible to update the package "setasign/fpdi" in "mpdf/mpdf" without breaking anything.

If it won't be possible I'll try to find other solution on my side.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
 		"ext-mbstring": "*",
 
 		"psr/log": "^1.0",
-		"setasign/fpdi": "1.6.*",
+		"setasign/fpdi": "^2.1",
 		"paragonie/random_compat": "^1.4|^2.0|9.99.99",
 		"myclabs/deep-copy": "^1.7"
 


lfluvisotto:
@finwe discard this PR, it'll need to adapt a lot "mpdf/mpdf" to use "setasign/fpdi": "^2.1"

I'll find out other way to solve this my case on my side.

finwe:
See this https://github.com/mpdf/mpdf/pull/899

Update to v2 is in the works.

GusAntoniassi:
@finwe are you able to estimate when v8 is going to be released? 

I'm currently using JanSlabon's PR as the repository on composer, but we're having some issues with GitHub's API limit when doing a `composer require`. We'd like to use MPDF for Packagist but we currently can't because of this issue.

Thanks in advance!

finwe:
@GusAntoniassi Please do not start new conversations in a closed and duplicate thread.

There is no v8 timeline ATM, but I guess I will perhaps postpone all other plans except incorporating the fpdi PR for the release. That should speed up the release enormously.

PR #973, opened 2019-02-25
Handle CSS fractions without a leading zero in font-size, like .5cm

### LINKED ISSUE #761: Leading zeroes in CSS numbers
### I found this bug / would like to have this new functionality

When using CSS values without leading zero like `width: .3cm;`, the rule is ignored. All modern browsers support numbers without leading zero (see http://hey.georgie.nu/leadingzero/) although there is no definite standardization.

We detected this bug when copy+pasting normal CSS code into mpdf and it did not work.

### This is a HTML/CSS code snippet I use
To reproduce, try the following HTML snippet:
```
<div style="font-size: .8cm;">This text shall be big</div>
<div style="font-size: .1cm;">This text shall be very small</div>
```

peterdevpl:
Resolves #761

### DIFF
--- a/src/CssManager.php
+++ b/src/CssManager.php
@@ -1051,13 +1051,6 @@ class CssManager
 			}
 		}
 
-		// Fix fractional sizes without leading zeros
-		foreach ($newprop as $k => $v) {
-			if (preg_match('/^\.[0-9]+[a-z%]*$/', $v)) {
-				$newprop[$k] = '0' . $v;
-			}
-		}
-
 		return $newprop;
 	}
 
--- a/src/CssManager.php
+++ b/src/CssManager.php
@@ -1051,6 +1051,13 @@ class CssManager
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


finwe:
Cool, thanks for the PR. Don't you think this would be better suited for SizeConverter class than the CssManager?

peterdevpl:
Your comment made me dig a bit deeper. It turned out that SizeConverter already supports fractions without a leading zero. Such values work with all CSS attributes except `font-size`. This is caused by a condition in `Mpdf::setCSS()`:

```
if (isset($arrayaux['FONT-SIZE'])) {
	$v = $arrayaux['FONT-SIZE'];
	if (is_numeric($v[0])) {
		/* value starting with a dot does not fall here */
	}
}
```

So I adjusted my fix a little bit and now it seems enough :)

finwe:
Squashed and merged to 8.0-RC1, thanks

PR #974, opened 2019-02-27
Fix #860: correctly initialize cell width for nested tables

### LINKED ISSUE #860: Nested tables broken since v7.1.0
Rendering nested tables does not resize containing table correctly. Following screenshot is from [example06 file](https://github.com/mpdf/mpdf-examples/blob/development/example06_tables_nested.php), mPDF versions 7.1.4 and 7.0.3 respectively.

shot_180903_085520

shot_180903_085533

Change in supported PHP version doesn't affect this.

@machour This is probably a result of the Tag refactoring, could you please have a look at this?

machour:
Fix issue introduced in https://github.com/mpdf/mpdf/commit/6cb5ddcaed56d0da49dea54d8245b01854f7faab by avoiding setting the cell width to 0 if it was undefined.

Using `null` instead of putting back the `if (isset(..)) { .. }` construct to avoid future confusion.

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -19174,7 +19174,7 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 					$c['padding']['L'] /= $k;
 
 					$c['maxs'] = isset($c['maxs']) ? $c['maxs'] /= $k : 0;
-					$c['w'] = isset($c['w']) ? $c['w'] /= $k : 0;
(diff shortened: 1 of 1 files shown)


finwe:
Wouldn't it make sense to make all keys previously enclosed in `isset` null as a default? To prevent the same behavior in other, unforseeable places?

machour:
Definitely makes sense. I was walking on eggs for tomorrow release so only touched what needed to be touch.

Adding another commit now

Issue #979, opened 2019-03-07
Bug with utf8 character


fglueck:
I found this bug:
MPDF produce the Error: Class Mpdf\Tag\ not found if I have the follow content inside any html:
💾
It doesn't matter if I use HTML Entity (hex/dec) or native.

I use:
$mpdf->WriteHTML($html, 0);

PHP Version 7.2 and MPDF v7.1.9

### This is a PHP code snippet I use

```
<?php
$html = <<<HTML
<html>
<body>
💾
</body>
</html>
HTML;
$mpdf->WriteHTML($html, 0);
```
Same problem with all character from this page:
https://www.fileformat.info/info/unicode/block/miscellaneous_symbols_and_pictographs/utf8test.htm

finwe:
Unable to reproduce with a phpunit test. Please provide additional information.

fglueck:
I found the bug, the method getTagInstance try to load any content.
The array after 
$a = preg_split('/<(.*?)>/ms', $html, -1, PREG_SPLIT_DELIM_CAPTURE);
Will contain like:
span
xyz <--- or any UTF8 Char
/span
getTagInstance now try's to load a class "xyz". This is not a valid TAG.
It seems to be a problem with different autoloaders, if one throws an exception if the class can't load...
I make a little, not perfect, workaround in getTagInstance. But I think it would be better to fix the preg_split... or is the content inside necessary?
It is also possible that:
```
		for ($i = 0; $i < $cnt; $i++) {
			$e = $a[$i];
			if ($i % 2 == 0) {
```
is not correct?, because why its try to load an empty class?
The array after regex_split look good:
```
array (
  0 => '',
  1 => 'html',
  2 => '',
  3 => 'body',
  4 => '🗿',
  5 => '/body',
  6 => '',
  7 => '/html',
  8 => '',
)
```

fglueck:
the calling line is:
```
	$tag = (isset($a2[0]) ? strtoupper($a2[0]) : '');
....
       $this->tag->OpenTag($tag, $attr, $a, $i); // mPDF 6
```
OpenTag will call with an empty $tag ('')

But the utf8 Symbol is no tag....

finwe:
This is alrigt and all, but I still cannot reproduce this. This leads me to believing there is some side condition to this. What are your Mpdf object config options?

fglueck:
```
   $settings = [
      'mode'          => $dest=='F' ? 'UTF-8-c' : 'UTF-8',
[...]

PR #980, opened 2019-03-07
Bugfix with utf8 char and Mpdf/Tag class not found

fglueck:


finwe:
Missing a test failing prior to this change.

finwe:
Related Issue https://github.com/mpdf/mpdf/issues/979

fglueck:
```
  private function getTagInstance($tag)   **<- if $tag is ''**
  {
        $className = self::getTagClassName($tag); **<- this generates 'Mpdf\Tag\'**
        if (class_exists($className)) { **<- this call the autoloader (query all pathes ... and if the autoloader can't find the file an error**
          return new $className(
            $this->mpdf,
            $this->cache,
            $this->cssManager,
            $this->form,
            $this->otl,
            $this->tableOfContents,
            $this->sizeConverter,
            $this->colorConverter,
            $this->imageProcessor,
            $this->languageToFont
          );
        }
      }
  }
```

finwe:
Not helpful. Create a unit test with a code sample actually reproducing this behaviour.

PR #989, opened 2019-03-17
Compound classes selector support

peterdevpl:
Examples: `.one.two`, `p.three.four.five`.

Other complex cases, like `#identifier.class` or `td:nth-child(even).special` are still not supported - they might need a lot of additional work.

Fixes mpdf#538


finwe:
Closed in 10ca235

kanujbhatnagar:
@peterdevpl, @finwe any chance when this change will be merged into the master branch? We're planning to use this feature for a critical release, but don't want to do it from the 'development' branch.

Thanks.

PR #1008, opened 2019-03-29
Add proxy support to curl

kakaska:


finwe:
Thanks. Can you please also add these two properties as keys to Mpdf\Config\ConfigVariables, so that they can be set while instantiating the mPDF object? Then, rebase the branch and squash the commits. Thanks!

src/Config/ConfigVariables.php
```diff
@@ -508,6 +508,8 @@ public function __construct()
 			'curlAllowUnsafeSslRequests' => false,
 			'curlCaCertificate' => '',
 			'curlTimeout' => 5,
+			'curlProxy' => '',
```
finwe: Use null as a default.

PR #1011, opened 2019-04-03
fix not existing array key ’s’ in _fixTableBorders

podolinek:
When complex html code  contains table with mixed css and html attributes, occured error "Undefined index: s", which not exists, but rendering still working when skip this.

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -21334,7 +21334,7 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 								} else {
 									$celladj = false;
 								}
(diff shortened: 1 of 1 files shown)


finwe:
Thanks for the PR. Please, add a unit test with HTML code reproducing this error.

PhoenixFnX:
Same issue here, can't give you a unit test either, our app is way too complex to do it. This commit fixes the problem. Any thought of drawback ?

mshYat:
@podolinek , @PhoenixFnX Try https://github.com/mpdf/mpdf/pull/1140 . See if that solves your problem.

podolinek:
All works. Thanks.

PR #1016, opened 2019-04-09
Performance improvements

ram4nd:
These might be considered as minor performance changes, but together they might make a dent.

- array_push(...) calls behaving as '$array []= ...'. ' []= ' works faster than invoking functions in PHP.
- ￼If you only want to determine if a particular needle occurs within haystack, use the faster and less memory intensive function strpos() instead.
- These cases can be refactored by using case-sensitive functions strstr(...)/strpos() for better performance because matched patterns does not contain any characters.
- substr() shorthand can be used instead of strlen().
- A 'substr()' call can behave as array access, but hiding some error-handling logic behind as well. Using array access is a preferred way, because you have to implicitly write (read as document) errors handling code.
- Cascade str_replace() calls when the function has been applied to a variable sequentially.
- Fix count function in loops.

### DIFF
--- a/src/Barcode.php
+++ b/src/Barcode.php
@@ -141,8 +141,14 @@ class Barcode
 
 	private function sanitizeCode($code)
 	{
-		$code = str_replace(chr(194) . chr(160), ' ', $code); // mPDF 5.3.95  (for utf-8 encoded)
-		$code = str_replace(chr(160), ' ', $code); // mPDF 5.3.95	(for win-1252)
+		$code = str_replace(
+		    array(
+		        chr(194) . chr(160), // mPDF 5.3.95  (for utf-8 encoded)
+                chr(160) // mPDF 5.3.95	(for win-1252)
+            ),
+            ' ',
+            $code
+        );
 
 		return $code;
 	}
--- a/src/Barcode.php
+++ b/src/Barcode.php
@@ -142,13 +142,13 @@ class Barcode
 	private function sanitizeCode($code)
 	{
 		$code = str_replace(
-		    array(
-		        chr(194) . chr(160), // mPDF 5.3.95  (for utf-8 encoded)
-                chr(160) // mPDF 5.3.95	(for win-1252)
-            ),
-            ' ',
-            $code
-        );
+			array(
+				chr(194) . chr(160), // mPDF 5.3.95  (for utf-8 encoded)
+				chr(160) // mPDF 5.3.95	(for win-1252)
+			),
+			' ',
+			$code
+		);
 
 		return $code;
 	}
--- a/src/Barcode.php
+++ b/src/Barcode.php
@@ -141,16 +141,10 @@ class Barcode
 
 	private function sanitizeCode($code)
 	{
-		$code = str_replace(
-			array(
-				chr(194) . chr(160), // mPDF 5.3.95  (for utf-8 encoded)
-				chr(160) // mPDF 5.3.95	(for win-1252)
-			),
-			' ',
-			$code
-		);
-
-		return $code;
+		return str_replace([
+			chr(194) . chr(160), // mPDF 5.3.95  (for utf-8 encoded)
+			chr(160) // mPDF 5.3.95	(for win-1252)
+		], ' ', $code);
 	}
 
 }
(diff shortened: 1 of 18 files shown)


src/Barcode.php
```diff
@@ -141,8 +141,14 @@ public function getBarcode($code, $type, $pr = 0.0)
 
 	private function sanitizeCode($code)
 	{
-		$code = str_replace(chr(194) . chr(160), ' ', $code); // mPDF 5.3.95  (for utf-8 encoded)
-		$code = str_replace(chr(160), ' ', $code); // mPDF 5.3.95	(for win-1252)
+		$code = str_replace(
+			array(
```
finwe: Short arrays, please.

src/Barcode/EanUpc.php
```diff
@@ -93,7 +93,7 @@ private function init($code, $length)
 			$invalidUpce = false;
 			if (($tmp == '000') or ($tmp == '100') or ($tmp == '200')) {
 				// Manufacturer code ends in 000, 100, or 200
-				$upceCode = substr($code, 2, 2) . substr($code, 9, 3) . substr($code, 4, 1);
+				$upceCode = substr($code, 2, 2) . substr($code, 9, 3) . $code[4];
```
finwe: Possible undefined key.

src/Barcode/EanUpc.php
```diff
@@ -106,16 +106,16 @@ private function init($code, $length)
 						$invalidUpce = true;
 					}
 				} else {
-					$tmp = substr($code, 6, 1);
+					$tmp = $code[6];
```
finwe: Possible undefined key.

src/Barcode/Postnet.php
```diff
@@ -65,14 +65,13 @@ private function init($code, $gapWidth, $planet)
 		$bararray = ['code' => $code, 'maxw' => 0, 'maxh' => 5, 'bcode' => []];
 
 		$k = 0;
-		$code = str_replace('-', '', $code);
-		$code = str_replace(' ', '', $code);
+		$code = str_replace(array('-', ' '), '', $code);
```
finwe: Short arrays.

src/Barcode/Postnet.php
```diff
@@ -65,14 +65,13 @@ private function init($code, $gapWidth, $planet)
...
 		$len = strlen($code);
 
 		// calculate checksum
 		$sum = 0;
 		for ($i = 0; $i < $len; ++$i) {
-			$sum += (int) $code[$i];
+			$sum += (int)$code[$i];
```
finwe: Please keep spaces after cast.

src/Form.php
```diff
@@ -382,7 +382,7 @@ function print_ob_select($objattr, $w, $h, $texto, $rtlalign, $k, $blockdir)
 			} // mPDF 5.3.37
 			$data = ['VAL' => [], 'OPT' => [], 'SEL' => [],];
 			if (isset($objattr['items'])) {
-				for ($i = 0; $i < count($objattr['items']); $i++) {
+				for ($i = 0, $iMax = count($objattr['items']); $i < $iMax; $i++) {
```
finwe: It is best to move the count call in front of the for loop entirely.

finwe:
You have still a 
> PHP Parse error:  syntax error, unexpected 'for' (T_FOR) in /home/finwe/www/_github/mpdf/repo/src/Otl.php on line 314

Also please fix the lint.

finwe:
More notices, with example61_new_mPDF_v6-0_features.php:

> PHP Notice:  Undefined offset: 474 in src/Otl.php on line 1017

When making a change this vast, I really recommend testing it out on the examples repository. The generate-examples.sh script there allows you to point mPDF root to a path so that you don't need to rely on local composer.

fulldecent:
@ram4nd Thank you for working on this. It is good to see some hand-tuning and learn some of these little habits.

fglueck:
Great job!

finwe:
@ram4nd Will you find time to finish this so it can be incorporated?

finwe:
Closing as stale. Feel free to reopen with fixes, rebased onto current development.

PR #1018, opened 2019-04-12
Rework the link reading code based on feedback from @JanSlabon

jakejackson1:
See https://github.com/mpdf/mpdf/commit/fc07ccb5ba39296376b5cc46ea659b9e7e704c39#commitcomment-33105075


JanSlabon:
This looks better/more stable now.

If you want to make it almost prefect you additionally need to take rotated pages into account.

Also the points in a Rectangle structure could be switched. In that case your width and height values will be negative.

jakejackson1:
This PR needs to be modified to account for imported pages in landscape mode, which appear to place the X/Y and width/height in reverse and causes display issues with the imported links. 

Also need to add back in:

```
		if ($annotations instanceof PdfIndirectObjectReference) {
			$annotations = PdfType::resolve($parser->getIndirectObject($annotations->value), $parser);
		}
```

To cover all bases.

finwe:
@jakejackson1 Will you find time to finish this? Can I help somehow?

jakejackson1:
@finwe I'm hoping to jump back on this next week. I need some simple test PDFs with links on rotated pages and also have the annotation rectangle points switched, so I can correctly fix and test for those use cases.

jakejackson1:
@finwe this should be ready to go!

PR #1023, opened 2019-04-26
Valid EAN UPC barcode and throw BarcodeException if is not valid.

PATROMO:
Because PHP throws warnings if `$code` has non-numeric chars:

```
PHP Warning:  A non-numeric value encountered in /vendor/mpdf/mpdf/src/Barcode/EanUpc.php on line 54
PHP Warning:  A non-numeric value encountered in /vendor/mpdf/mpdf/src/Barcode/EanUpc.php on line 62
```

### DIFF
--- a/src/Barcode/EanUpc.php
+++ b/src/Barcode/EanUpc.php
@@ -35,6 +35,10 @@ class EanUpc extends \Mpdf\Barcode\AbstractBarcode implements \Mpdf\Barcode\Barc
 	 */
 	private function init($code, $length)
 	{
+        if (preg_match('/[\D]+/', $code)) {
+            throw new \Mpdf\Barcode\BarcodeException('Invalid EAN UPC barcode value');
+        }
+
 		$upce = false;
 		$checkdigit = false;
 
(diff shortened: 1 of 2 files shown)


src/Barcode/EanUpc.php
```diff
@@ -35,6 +35,10 @@ public function __construct($code, $length, $leftMargin, $rightMargin, $xDim, $b
 	 */
 	private function init($code, $length)
 	{
+		if (preg_match('/[\D]+/', $code)) {
```
finwe: This will allow any string with numbers. EAN/UPC barcodes can use digits only.

src/Barcode/EanUpc.php
```diff
@@ -35,6 +35,10 @@ public function __construct($code, $length, $leftMargin, $rightMargin, $xDim, $b
...
+			throw new \Mpdf\Barcode\BarcodeException('Invalid EAN UPC barcode value');
```
finwe: Throwing an exception would mean this should be postponed for 9.0. What happens when the exception is not thrown?

finwe:
Never mind the review. Thanks.

PR #1026, opened 2019-04-30
Adding the check for JPEG SOF header 0xFF 0xC1 (extended) - This is s…

jamiejones85:
The ImageProcessor is only looking for 2 out of the 4 possible SOF headers.

- FFC0 (baseline) - This is the usual mode chosen for photos and encodes fully specified DCT blocks in groupings depending on the color/subsample options chosen
- FFC1 (extended) - This is similar to baseline, but has more than 8-bits per color stimulus
- FFC2 (progressive) - This mode is often found on web pages to allow the image to load progressively as the data is received. Each "scan" of the image progressively defines more coefficients of the DCT blocks until they're fully defined. This effectively provides more and more detail as more scans are decoded
- FFC3 (lossless) - This mode uses a simple Huffman encoding to losslessly encode the image. The only place I've seen this used is on 16-bit grayscale DICOM medical images

Source: https://stackoverflow.com/questions/35427283/get-width-and-height-from-jpeg-without-0xff-0xc0 comment by BitBank.

I've added 0xFF 0xC1 and this allows my test jpeg to be rendered in the pdf where previously I was getting 'Error parsing JPG header'


src/Image/ImageProcessor.php
```diff
@@ -282,7 +282,6 @@ public function getImage(&$file, $firsttime = true, $allowvector = true, $orig_s
 
 			return $info;
 		}
-
```
finwe: Please keep whitespace as were.

src/Image/ImageProcessor.php
```diff
@@ -1312,14 +1311,13 @@ private function jpgHeaderFromString(&$data)
 		$p = 4;
 		$p += $this->twoBytesToInt(substr($data, $p, 2)); // Length of initial marker block
 		$marker = substr($data, $p, 2);
-
-		while ($marker !== chr(255) . chr(192) && $marker !== chr(255) . chr(194) && $p < strlen($data)) {
-			// Start of frame marker (FFC0) or (FFC2) mPDF 4.4.004
+		while ($marker !== chr(255) . chr(192) && $marker !== chr(255) . chr(194)  && $marker !== chr(255) . chr(193) && $p < strlen($data)) {
+			// Start of frame marker (FFC0) (FFC1) or (FFC2) mPDF 4.4.004
```
finwe: Please remove the mPDF version from the comment as it is misleading now.

src/Image/ImageProcessor.php
```diff
@@ -1312,14 +1311,13 @@ private function jpgHeaderFromString(&$data)
 		$p = 4;
 		$p += $this->twoBytesToInt(substr($data, $p, 2)); // Length of initial marker block
 		$marker = substr($data, $p, 2);
-
```
finwe: Removed whitespace.

finwe:
Also no trailing whitespace, please. And squash commits to one and force push, please.

jamiejones85:
Have I missed anything? I'm not sure how to squash commits, I've not done that before.

finwe:
You've left two tabs on an otherwise empty line 285.

Never mind, I've fixed that for you in ee370f8, squashed the commits and added a changelog entry.

Closing this.

PR #1039, opened 2019-05-16
fix mpdf overwrites setasign/fpdi/fpditrait cleanup method

wecler:
Reason:
cleanup method from setasign/Fpdi/FpdiTrait is being overwritten by Mpdf.
Fix:
Rename cleanup method from Mpdf to resetEncoding.


finwe:
The name `resetEncoding` is too specific. The method is meant to reset various global settings, not just encoding.

Also, as this is a breaking change, this cannot be incorporated until v9.

PR #1043, opened 2019-06-06
Make php extensions optional

alexander-schranz:


finwe:
This is an absolutely insufficient change.

For the GD extension to be removed from composer, I would insist on adding checks to code using those extensions explaining why the code is not running and that an installation of an extension is needed - such in https://github.com/mpdf/mpdf/blob/development/src/Barcode/Imb.php#L46

mPDF without mbstring won't run at all (se an exception thrown in Mpdf.php:9493 during mPDF instantiation) - removing mbstring from composer doesn't make sense.

finwe:
What is your reasoning for this change, anyway?

alexander-schranz:
For me it looks gd is correctly handled in the code or is there any place where it isn't?

https://github.com/mpdf/mpdf/blob/ee370f89a21956ec0ea3e38c0678ced491c6b56a/src/Image/ImageProcessor.php#L555-L558

> mPDF without mbstring won't run at all (se an exception thrown in Mpdf.php:9493 during mPDF instantiation) - removing mbstring from composer doesn't make sense.

thats why the mb polyfill is added, which will use the extension when available.

alexander-schranz:
> What is your reasoning for this change, anyway?

@finwe no need to have the extensions installed for basic usage

finwe:
I think GD and mbstring are present in most of PHP installations. I still rather not introduce another dependencies. Thanks for the idea, yet I'm going to close this and won't incorporate this, at least for now.

guilliamxavier:
Just so you know: I have to do text-only HTML to PDF conversion on a PHP system where the GD extension is not installed (and won't be, because it is not supposed to handle images), and `composer require mpdf/mpdf` fails with error `mpdf/mpdf v8.0.7 requires ext-gd * -> the requested PHP extension gd is missing from your system`, so I won't use mPDF but some other package like Dompdf (which requires MBString and DOM [both installed], but only *suggests* GD).

finwe:
Good for you, then.

This doesn't change the fact that this PR was insufficient - with this change and GD library disabled, two examples from the example repository failed ungracefully and one got stuck in an infinite loop.

guilliamxavier:
@finwe: Ah, not great indeed... thanks for checking anyway (which makes me wonder if the automated test suite could be improved so that this PR would *not* be green on Travis CI?)

finwe:
yes, but not quite easily https://github.com/mpdf/mpdf/issues/1303

PR #1044, opened 2019-06-07
fix: Display external links on imported PDFs

pecan987:
Reason:
Bad condition in getImportedExternalPageLinks() caused ignoring some links in imported pages.


finwe:
Please, add a code sample failing without this change, add a unit test with a sample influenced by this change, so we can review this correctly.

PR #1045, opened 2019-06-13
fopen() vs is_readable()

fglueck:


src/Tag.php
```diff
@@ -108,21 +108,23 @@ public function __construct(
...
-				$this->otl,
-				$this->tableOfContents,
-				$this->sizeConverter,
-				$this->colorConverter,
-				$this->imageProcessor,
-				$this->languageToFont
-			);
-		}
+	  if (preg_match('/^[a-zA-Z]+$/', $tag)) {
+	    $className = self::getTagClassName($tag);
```
Thoulah: Please keep indent to tabs: https://github.com/mpdf/mpdf/blob/development/ruleset.xml

finwe:
Thanks. Please, fix the CI, squash middle two commits, and add a unit test reproducing the faulty behaviour for the change with utf8 tag characters.

For changes to be incorporated more quickly, separate different concerns of code to their own pull requests.

src/Mpdf.php
```diff
@@ -9407,7 +9407,7 @@ function Output($name = '', $dest = '')
 						header('Content-Length: ' . strlen($this->buffer));
 					}
 
-					header('Content-disposition: inline; filename="' . $name . '"');
+					header('Content-Disposition: attachment; filename="' . $name . '"; filename*=utf-8\' \'' . rawurlencode($name));
```
Klap-in: Could you give an explanation why this should be changed from inline to attachment?

finwe:
I am going to close this as too broad and stale. @fglueck please, stick with a single concern per PR and always include a unit test that actually covers changes in the code.

Issue #701, opened 2018-03-27
tempDir setting is idioatic


worikgh:
In ConfigVariables.php 
`'tempDir' => __DIR__ . '/../../tmp',`
Why should this directory exist?  The default should be to use `sys_get_temp_dir`
Passing ['tempDir' => sys_get_temp_dir()] as a argument to the sonstructor gets around it

Klap-in:
This definition is in the `ConfigVariables.php` file, which lives in: https://github.com/mpdf/mpdf/blob/development/src/Config/ConfigVariables.php

So two levels up, and you are in the root directory. 
In this root directory lives the `tmp/` folder.
See: https://github.com/mpdf/mpdf/tree/development/tmp

This folder exists to ensure it exists somewhere.

Of course it is better to define it at a more suitable location.
Please see the mpdf manual for details:
https://mpdf.github.io/installation-setup/folders-for-temporary-files.html

worikgh:
`mPDF is pre-configured to use <path to mpdf>/tmp as a directory to write temporary files (mainly for images). Write permissions must be set for read/write access for the tmp directory.`
Why?  What is wrong with `sys_get_temp_dir`?.

finwe:
Feel free to propose a PR with the change.

finwe:
System wide temporary directory presents potential collisions of multiple mPDF instances used by different system users, especially on shared hosts. Custom temporary directory for each installation is safer.

madorin:
Using /tmp inside mPdf lib, where sources are located, is not the best idea from security point of view.
Most people will assign full rights to entire source folder to apache/nginx user, normally it should be RO.
Also you have to play with the folder rights each time on site update.
Collisions are solved somehow. This is proved by 99.5% of apps, that uses system temporary directory.

Klap-in:
It is recommended to use another folder, according to docs...: https://mpdf.github.io/installation-setup/installation-v7-x.html and https://mpdf.github.io/installation-setup/folders-for-temporary-files.html)

finwe:
The main argument against `sys_get_temp_dir()` (actually a no-go) in a PHP application is `open_basedir` restrictions applied on many shared hosts. PHP is not able to write and read in `/tmp`.

The default `tempDir` setting is for sake of backwards compatibility.

Its value can be changed on mPDF instantiation to any suitable value by user.

It is possible `tempDir` will not have a default value and will be required to be set by user in a future major version.

KazuAlex:
[...]

PR #1047, opened 2019-06-14
use system temp dir for default tempDir value

KazuAlex:
If the vendor dir is not writable for any reason, we can't create tempDir folder to it.
While the system temp dir assume to be always writable


finwe:
[Been there](https://github.com/mpdf/mpdf/issues/701#issuecomment-394724196). You can always set a custom writable value.

PR #1056, opened 2019-07-15
security, open require in COLLATION-GROUP

timglabisch:
currently it is possible to misuse `COLLATION-GROUP` to `require` arbitrary php files.
This can be dangerous if you accept html from third party.

Just have a look at the test. The current master/development would fail because it cant locate FOOBAR.php.

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -23691,6 +23691,20 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 		$this->Reference[] = ['t' => $txta . ' - see ' . $txtb, 'p' => []];
 	}
 
+	private function filesInDir($directory)
+	{
+		$files = [];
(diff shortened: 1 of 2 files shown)


finwe:
Thanks for the catch and the fix!

I'm not sure about using the `glob` function which may be disabled on some shared hosts. 

Maybe a character whitelist (`a-zA-Z\(\)_`) for the collation and subsequent file_exists call would be a cleaner solution?

timglabisch:
@finwe from security perspective i prefer whitelisting over blacklisting. May `DirectoryIterator` ?

finwe:
DirectoryIterator would certainly be better.

timglabisch:
alright, i am going to change this in the next few hours.

timglabisch:
@finwe done.

finwe:
Thanks, can you please squash the commits to one?

timglabisch:
@finwe done.

Issue #949, opened 2019-01-08
Insecure PHP deserialization through phar:// wrapper.


ByQwert:
Hello, 

During bug bounty hunting I met one of the old version of yours library and decided to test it for known vulnerabilities, namely PHP deserialization through _phar://_ wrapper that was discovered independently by @s-n-t and @orangetw.

[Presentation Slides by Sam Thomas ](https://i.blackhat.com/us-18/Thu-August-9/us-18-Thomas-Its-A-PHP-Unserialization-Vulnerability-Jim-But-Not-As-We-Know-It.pdf)
[White Paper by Sam Thomas](https://i.blackhat.com/us-18/Thu-August-9/us-18-Thomas-Its-A-PHP-Unserialization-Vulnerability-Jim-But-Not-As-We-Know-It-wp.pdf)
[CTF challenge by Orange Tsai](https://github.com/orangetw/My-CTF-Web-Challenges/blob/master/README.md#babyh-master-php-2017)

So, after some tests, looks like yours library has similar issue as [TCPDF](https://github.com/tecnickcom/TCPDF) library.

Method _getImage()_ of _Image/ImageProcessor_ class pass value of _src_ attribute of _img_ tag to **fopen()** function, what can lead to PHP deserialization if value contains _phar://_ wrapper.
https://github.com/mpdf/mpdf/blob/development/src/Image/ImageProcessor.php#L215

For proof of concept I create class with ___destruct()_ method (based on GuzzleHttp deserialization chain) and use [phpggc](https://github.com/ambionics/phpggc) library to create image, which contains phar metadata. Phpggc repo also provide dozens of popular deserialization chains for popular frameworks and libraries.

Source code of test class:
```
<?php

namespace TestLib;

class VulnerableClass {

    private $file;
    private $content;

    public function __construct($file, $content) {
        $this->file = $file;
        $this->content = $content;
    }

    /* ... */

    public function __destruct() {
        $this->save($this->file, $this->content);
    }

    public function save($file, $content) {
        file_put_contents($file, $content);
    }
}

?>
```

Source code of trigger script:
```
<?php

require('lib/class.php');

require_once __DIR__ . '/../vendor/autoload.php';

$input = $_GET['input'];

if (!$input) {
    $input = "";
}

$mpdf = new \Mpdf\Mpdf();
$mpdf->WriteHTML($input);
$mpdf->Output();

?>
```
[...]

PR #1057, opened 2019-07-15
Filesystem wrapper for sandboxing

timglabisch:
currently it's nearly impossible to restrict / sandbox mpdf.
especially if you accept third party html this could end up in security flaws.

Some Examples:
- https://github.com/mpdf/mpdf/issues/949
- Timeouts in Curl -> Images, CSSManager
- Loading local / remote Images

from security perspective it would be great if it would be possible to restrict mpdf's io.
this PR is the first step to this direction.

i created a new class for handling (hopefully) all io releated stuff. In Later PR's we could create an interface for this class and allow users so restrict/sandbox IO functionality.

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -15,18 +15,18 @@ class Cache
 
 	public function __construct($basePath, FileSystem $fileSystem, $cleanupInterval = 3600)
 	{
+        $this->basePath = $basePath;
+        $this->fileSystem = $fileSystem;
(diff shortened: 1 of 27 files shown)


timglabisch:
anything I can do to make this happen?

finwe:
Closing this in favor of the LocalContentLoader introduced in upcoming v8.1.0

finwe:
(Other services will be allowed to be replaced along the way, possibly including cache etc.)

PR #1086, opened 2019-09-10
Add support for PHP 7.4

snapshotpl:


snapshotpl:
I leave this as is. I have fixed all deprecations and issues what I can.

There is only 5 issues to fix. I see problem with tests and also implementation, so I don't know how to figure it out.

Important is that this issues are probably in every php versions however old phps allow them to live without any error.

finwe:
Thanks for the effort, if you are not going to finish this, I am going to close this and finish my local compatibility branch where I got to the same point, basically.

PR #1100, opened 2019-10-07
Fixed some Array-style access of non-arrays problems due a PHP 7.4

spagu:
My system was reporting problems generating PDF files in PHP7.4 like notices for PHP:

> Notice: Trying to access array offset on value of type int in /vendor/mpdf/mpdf/src/Mpdf.php on line 21471

This is due [an incompatibility from PHP7.4 ](https://www.php.net/manual/en/migration74.incompatible.php)

I did fix problems that affect only me, and it works but I am not sure if this is as it should be.


finwe:
This is handled in https://github.com/mpdf/mpdf/pull/1091

Issue #253, opened 2016-10-27
Error parsing image file exception on HTTP errors with cURL


bcarroll2k3:
I have an issue that is driving me insane, hopefully someone can shed some light.  I have pages in a WordPress environment that generate PDFs.  One is in a directory located at the root of the WP filesystem and is called directly (no WP URL rewrite).  The other is served from WP and the code for PDF generation is located in functions.php for the theme.

The page located in the directory at the WP root works fine.  The PDF generated from functions.php, which is nearly identical to the other page, constantly throws this error:

```
mPDF error: IMAGE Error (http://dev.xxxx.com/wp-content/themes/rw/img/logo_new.gif): Error parsing image file - image type not recognised, and not supported by GD imagecreate
```

The html that's being rendered is identical.  I'm using `WriteFixedPosHTML` to add an image to the document.  The page that works defines this image as:

```
$img = '<img src="../wp-content/themes/rw/img/logo_new.gif" style="width: 145px;">';
```

The code in functions.php, which doesn't work, defines the image as:

```
$img = '<img src="http://dev.xxxx.com/wp-content/themes/rw/img/logo_new.gif" style="width: 145px;">';
```

I have to define it this way otherwise I get the error regardless.  I have tried the following:
- Setting src local to function.php:  src="img/logo_new.gif", results in

```
mPDF error: IMAGE Error (http://dev.xxxx.com/img/logo_new.gif): Error parsing image file - image type not recognised, and not supported by GD imagecreate
```
- Setting src to be src="../rw/img/log_new.gif", results in

```
mPDF error: IMAGE Error (http://dev.xxxx.com/img/logo_new.gif): Error parsing image file - image type not recognised, and not supported by GD imagecreate
```

note the path hasn't changed
- Setting src to be src="http://dev.xxxx.com/wp-content/themes/rw/img/logo_new.gif" results in

```
mPDF error: IMAGE Error (http://dev.xxxx.com/wp-content/themes/rw/img/logo_new.gif): Error parsing image file - image type not recognised, and not supported by GD imagecreate
```

So, for the life of me I don't understand why
1. One page works with relative pathing and the other doesn't (???)
2. The page that doesn't work doesn't seem to recognize the image as being valid.  I can browse to http://dev.xxxx.com/wp-content/themes/rw/img/logo_new.gif just fine.

The php code used to generate pdfs from both pages is identical

```
[...]

PR #1102, opened 2019-10-08
Improve debugging

ribeirobreno:
Improves debugging remote content for mPDF by:
- Issuing an exception if debug is enabled and a remote content request ends with:
-- a cURL error
-- a HTTP status other than 200
- Including a log of HTTP status other than 200

This should make it easier to understand several causes of #253 by just enabling [debug mode](https://mpdf.github.io/reference/mpdf-variables/debug.html).


ribeirobreno:
Closing and reopening to trigger a build. Seems like it failed due to an environment issue.

PR #1108, opened 2019-10-17
Add page rotation method

### LINKED ISSUE #902: rotating pages
Would it be possible to introduce an option to custom rotate single pages by a multiple of 90 degrees? This would be helpful in order to prepare a PDF for mobile duplex printing, where the flipping edge cannot be specified. Rotating the even pages upside down does the trick - but I haven't found any option to do this within MPDF.

Aragur:
This PR adds a function `RotatePage(str::degree)` which allows a user to rotate specific pages.  
This has to be injecteted before `Output()` or `AddPage()` is called.

This also fixes #902

### DIFF
--- a/src/Writer/PageWriter.php
+++ b/src/Writer/PageWriter.php
@@ -144,10 +144,11 @@ final class PageWriter
 			$this->writer->object();
 			$this->writer->write('<</Type /Page');
 			$this->writer->write('/Parent 1 0 R');
-			if(array_key_exists($n, $this->mpdf->rotations))
+			if (array_key_exists($n, $this->mpdf->rotations)) {
 				$this->writer->write('/Rotate ' . $this->mpdf->rotations[$n]);
-			else 
+			} else {
 				$this->writer->write('/Rotate 0');
+			}
 
 			if (isset($this->mpdf->OrientationChanges[$n])) {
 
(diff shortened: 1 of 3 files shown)


soberdor:
Thank you very much for making this work, I already used your changes in my project!

One suggestion, the method could allow for specifying the page - if the decision to rotate is taken at a later stage.

I adapted it like this:

```
function RotatePage($r, $page=false)
{
	if ($page === false) $page=$this->page;
	$this->rotations[$page] = $r;
}	
```

adamgp:
Any hope of having this simple feature merged in anytime soon?

michaljusiega:
cc @finwe I think this feature is important not only for me.
Can we merge this PR ?

finwe:
There is a unit test using this new code missing. After the test is added, I'll be happy to merge this. Also, documentation update has to be prepared.

src/Writer/PageWriter.php
```diff
@@ -144,6 +144,11 @@ public function writePages() // _putpages
 			$this->writer->object();
 			$this->writer->write('<</Type /Page');
 			$this->writer->write('/Parent 1 0 R');
+			if (array_key_exists($n, $this->mpdf->rotations)) {
+				$this->writer->write('/Rotate ' . $this->mpdf->rotations[$n]);
+			} else {
+				$this->writer->write('/Rotate 0');
```
finwe, line 150: Is this 0 rotation necessary or can it be omitted?

adamgp: Omitted.

fabianwohlfart:
+1
This would add a rotation to AddPage / AddPageByArray like in FPDF?

finwe:

Issue #1134, opened 2019-12-04
Table border style bug. Undefined index: s. Undefined index: w.


mshYat:
### I found this bug
PDF failed to generate.
```
Notice: Undefined index: s in mpdf\src\Tag\Td.php on line 297
Notice: Undefined index: w in mpdf\src\Mpdf.php on line 19378
Notice: Undefined index: w in mpdf\src\Mpdf.php on line 20181
Notice: Undefined index: w in mpdf\src\Mpdf.php on line 22749

Notice: Undefined index: s in mpdf\src\Tag\Td.php on line 297
Notice: Undefined index: w in mpdf\src\Mpdf.php on line 19378
Notice: Undefined index: w in mpdf\src\Mpdf.php on line 20181
Notice: Undefined index: w in mpdf\src\Mpdf.php on line 22749
```

### This is mPDF and PHP version and environment (server/fpm/cli etc) I am using
I am using: mpdf v8.0.4, PHP 7.3

### This is a PHP code snippet I use

```
<?php
$html = '
<table><tbody><tr><td style="border-left-color: rgb(207, 207, 207);">test</td></tr></tbody></table>
<table><tbody><tr><td style="border-left-color: rgb(207, 207, 207);">test</td></tr></tbody></table>
';

$mpdf= new \Mpdf\Mpdf();
$mpdf->WriteHTML($html);
$mpdf->Output();
```

PR #1140, opened 2019-12-13
Fix 1134 issue

mshYat:
[Issue 1134](https://github.com/mpdf/mpdf/issues/1134)


finwe:
Thanks, please add a unit test demonstrating this fix.

PR #1160, opened 2020-01-29
Fix TTCfontID assignment.

bensmithurst:
Previous code was incorrectly assigning the result of isset() rather
than the value itself.

Apologies for the lack of tests for this, I haven't had time to do that yet but I wanted to submit this fix before I forget about it, as it might benefit others scratching their heads over this problem.


finwe:

PR #1166, opened 2020-02-10
added AddFontData method and implemented dry

ajayexpert:
If we have the AddFontData method that makes adding font very easy with the use https://mpdf.github.io/reference/mpdf-functions/addfontdirectory.html method 

```php
$mpdf = new Mpdf([
        ...
    ]);

$mpdf->AddFontDirectory("path/of/custom/fonts");
$mpdf->AddFontData([
        'frutiger' => [
            'R' => 'Frutiger-Normal.ttf',
            'I' => 'FrutigerObl-Normal.ttf',
        ],
        'customfontbyme' => [
            'R' => "customfontbyme.ttf",
        ]
    ]);
$mpdf->SetDefaultFont("frutiger");
```

<br>
Instead of doing this -> 

```php

$defaultConfig = (new Mpdf\Config\ConfigVariables())->getDefaults();
$fontDirs = $defaultConfig['fontDir'];

$defaultFontConfig = (new Mpdf\Config\FontVariables())->getDefaults();
$fontData = $defaultFontConfig['fontdata'];

$mpdf = new \Mpdf\Mpdf([
    'fontDir' => array_merge($fontDirs, [
        __DIR__ . '/custom/font/directory',
    ]),
    'fontdata' => $fontData + [
        'frutiger' => [
            'R' => 'Frutiger-Normal.ttf',
            'I' => 'FrutigerObl-Normal.ttf',
        ]
    ],
    'default_font' => 'frutiger'
]);
```

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -1320,21 +1320,7 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 
 		// Available fonts
 		$this->available_unifonts = [];
-		foreach ($this->fontdata as $f => $fs) {
-			if (isset($fs['R']) && $fs['R']) {
-				$this->available_unifonts[] = $f;
-			}
-			if (isset($fs['B']) && $fs['B']) {
-				$this->available_unifonts[] = $f . 'B';
-			}
-			if (isset($fs['I']) && $fs['I']) {
-				$this->available_unifonts[] = $f . 'I';
-			}
-			if (isset($fs['BI']) && $fs['BI']) {
-				$this->available_unifonts[] = $f . 'BI';
-			}
-		}
-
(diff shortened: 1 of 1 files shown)


finwe:
Thanks for the PR, but using one method (mostly) does not make sense without using the other.

Furthermore, this will be hopefully made obsolete with (a long overdue) font subpackages change, see https://github.com/mpdf/mpdf/pull/841.

finwe:
Meaning, let's do a proper cleanup instead of introducing "temporary" methods that we already know will be deprecated in the next major version.

ajayexpert:
Are we going to introduce with new way of adding custom font in 8.0 or 9.0

finwe:
It's a breaking change, so in 9.0, as the referenced PR milestone suggests.

PR #1189, opened 2020-03-25
Fix japanise french and other chars in file name in ie11

sergey347:


finwe:
Thanks for a PR. I will not encorporate it though, it is not the library job to hack browsers' bugs. You can hack this in your own code.

Issue #1060, opened 2019-07-26
Undefined index: inherit when using Bootstrap 4 and TD


shahimclt:
### I found this bug / would like to have this new functionality

When using with a 'td' tag and bootstrap 4, I get the following error:

> Notice: Undefined index: inherit in C:\xampp\htdocs\engage_web\vendor\mpdf\mpdf\src\Tag\Td.php on line 208

### This is mPDF and PHP version and environment (server/fpm/cli etc) I am using

PHP version: 7
mPDF version: 8.0.2

### This is a PHP code snippet I use

    $mpdf->WriteHTML($this->load->view('survey/surveyReport_printable', $data));

Thats all I've been able to narrow down. I would like some workarounds and/or some pointers to help me fix it myself.

ssieb:
Get the html output of
$this->load->view('survey/surveyReport_printable', $data)
Try viewing it in a web browser and try to reduce it to the minimum required to cause the error.  Particularly look for some CSS with "inherit" in it.  The developers can't do anything with no data.

GabrielCNoble:
Linked is a simple example that triggers this problem.

[https://pastebin.com/r96ZL04b](url)

GabrielCNoble:
This is enough to trigger the problem.

[pastebin.com/VaALc0Tz](url)

Servietteski:
Bootstrap isn't supported by MPDF, it's annoying. Same issue

thallesdaniell:
I detected this and decided to extract the bootstrap for the style tag and tested it until I reached that style and removing it worked for me
```
    th {
        text-align: inherit
    }
```

MarkVaughn:
same breaking when parsing random emails containing `text-align: initial`

finwe:
One-line comments without meaningful code example are not helpful. Even more so when the original issue is aslo missing full code example.

Please create separate issues with reproducible code samples to faster reproduce and therefore fix the issue.

MarkVaughn:
From: 
https://developer.mozilla.org/en-US/docs/Web/CSS/text-align
Here are possible values for text-align
```
/* Keyword values */
text-align: left;
text-align: right;
text-align: center;
text-align: justify;
text-align: justify-all;
text-align: start;
text-align: end;
text-align: match-parent;

/* Character-based alignment in a table column */
text-align: ".";
text-align: "." center;

/* Block alignment values (Non-standard syntax) */
text-align: -moz-center;
text-align: -webkit-center;

/* Global values */
[...]

PR #1194, opened 2020-04-08
Created Align Helper Method

MarkVaughn:
Instead of accessing the value via `self::ALIGN[strtolower('foo')` `$this->getAlign('foo')` 
Replacing 26 occurrences  and also preventing an exception being thrown for align values we don't know (initial, inherit, start, end)

Related #847 &  #1060 

Currently  \Mpdf\Tag\Tag only supports the values defined in the const which is incomplete
```
	const ALIGN = [
		'left' => 'L',
		'center' => 'C',
		'right' => 'R',
		'top' => 'T',
		'text-top' => 'TT',
		'middle' => 'M',
		'baseline' => 'BS',
		'bottom' => 'B',
		'text-bottom' => 'TB',
		'justify' => 'J'
	];
```
From: 
https://developer.mozilla.org/en-US/docs/Web/CSS/text-align
Here are possible values for text-align
```
/* Keyword values */
text-align: left;
text-align: right;
text-align: center;
text-align: justify;
text-align: justify-all;
text-align: start;
text-align: end;
text-align: match-parent;

/* Character-based alignment in a table column */
text-align: ".";
text-align: "." center;

/* Block alignment values (Non-standard syntax) */
text-align: -moz-center;
text-align: -webkit-center;

/* Global values */
text-align: inherit;
text-align: initial;
text-align: unset;
```

So this helper prevents the notice from being thrown when `self::ALIGN[$foo]` is not set

### DIFF
--- a/src/Tag/Columns.php
+++ b/src/Tag/Columns.php
@@ -47,7 +47,7 @@ class Columns extends Tag
 				if ($attr['VALIGN'] === 'J') {
 					$valign = 'J';
 				} else {
-					$valign = self::ALIGN[$attr['VALIGN']];
+					$valign = $this->getAlign($attr['VALIGN']);
 				}
 			} else {
 				$valign = '';
--- a/src/Tag/BarCode.php
+++ b/src/Tag/BarCode.php
@@ -151,7 +151,7 @@ class BarCode extends Tag
 			}
 
 			if (isset($properties['VERTICAL-ALIGN'])) {
-				$objattr['vertical-align'] = self::ALIGN[strtolower($properties['VERTICAL-ALIGN'])];
+				$objattr['vertical-align'] = $this->getAlign($properties['VERTICAL-ALIGN']);
 			}
 			if (isset($properties['COLOR']) && $properties['COLOR'] != '') {
 				$objattr['color'] = $this->colorConverter->convert($properties['COLOR'], $this->mpdf->PDFAXwarnings);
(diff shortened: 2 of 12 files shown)


src/Tag/Columns.php
```diff
@@ -47,7 +47,7 @@ public function open($attr, &$ahtml, &$ihtml)
 				if ($attr['VALIGN'] === 'J') {
 					$valign = 'J';
 				} else {
-					$valign = self::ALIGN[$attr['VALIGN']];
```
finwe, line 50: The attr uses uppercase here. It is possible that the case will mismatch.

MarkVaughn: Noted

MarkVaughn:
Not sure why Travis is not reporting to Github. Here's the build
https://travis-ci.org/github/mpdf/mpdf/builds/672602614

src/Tag/Columns.php
MarkVaughn: Wait that doesn't make sense -
 `self::ALIGN[$attr['VALIGN']];` if $attr['VALIGN'] is upper case like `CENTER` then `self::ALIGN['CENTER']` breaks

MarkVaughn:
@finwe I added a test which failed before the helper fix, is this good to go?

finwe:
I have to look at this, please give me some more time.

finwe:
Merged into development, thanks.

PR #1204, opened 2020-04-14
Fix non http URI schemes

MarkVaughn:
Addresses Exception:
```
ErrorException is_file(): Unable to find the wrapper "chrome-extension" - did you forget to enable it when you configured PHP? 
    /var/www/foo/vendor/mpdf/mpdf/src/Mpdf.php:11353 Illuminate\Foundation\Bootstrap\HandleExceptions::handleError
    [internal] is_file
    /var/www/foo/vendor/mpdf/mpdf/src/Mpdf.php:11353 Mpdf\Mpdf::GetFullPath
    /var/www/foo/vendor/mpdf/mpdf/src/Mpdf.php:13591 Mpdf\Mpdf::WriteHTML
```

There are plenty of URI schemes other than fax, tel & mailto to ignore
https://en.wikipedia.org/wiki/List_of_URI_schemes & https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

I created a pattern that ignores all  non-HTTP schemes, [here it is in action ](https://regex101.com/r/4RqJMZ/1 )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ mPDF 8.0.x
 * Fixed skipping ordered list numbering with page-break-inside: avoid (#339)
 * Compound classes selector support, like `.one.two` or `div.message.special` (#538, @peterdevpl)
 * Fixed CMYK colors in text-shadow (#1115, @lexilya)
+* Skip non supported wrappers when resolving paths (#1204, @MarkVaughn)
 
 mPDF 8.0.0
 ===========================
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -11425,7 +11425,10 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 			return;
 		}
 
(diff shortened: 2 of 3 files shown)


finwe:
Thanks for a PR.

This will ignore many more schemes that could/should be linked, such as ftp://, file://, smb:// etc.

Can you prepare a less restrictive implementation? Maybe using suitable wrappers mentioned in PHP docs? https://www.php.net/manual/en/wrappers.php

What about enabling custom wrappers by user request? 

Also please add a changelog entry, squash commits and use more meaningful commit message.

MarkVaughn:
OK, with my change you can now use custom  wrappers,
example on my system
```
>>> stream_get_wrappers()
=> [
     "https",
     "ftps",
     "compress.zlib",
     "compress.bzip2",
     "php",
     "file",
     "glob",
     "data",
     "http",
     "ftp",
     "phar",
     "zip",
   ]

```

finwe:
Great! Please fix lint issues.

finwe:
Merged in 67aafe7, thanks!

Issue #450, opened 2017-07-25
All my svg images get filled with black color


Samz3n:
Hi all,

When I try to embed an image-tag with a svg file
`<img src="/path-to-image/logo.svg">`

It will get filled with black color when exported to PDF.

The fill-color get overruled by the construct function in "/classes/svg.php"
`$this->svg_style = array(
	array(
		'fill' => 'black',
		...
	)
);`

How can I avoid it from happen? My logo has multiple color.

Samz3n:
I found a working solution.
It is working when I save my .svg files from Inkscape. 
More complex vector files won't work when saving from Illustrator. 

- I tried testing it a little bit more. 
I succeeded to save to working logos in right colors from Illustrator (however I had to modify the code a little bit for one of them).

I also have a more complex vector image with a lot of shapes. I cannot save a working .svg file for this vector from Illustrator but it works when I save from Inkscape.

finwe:
can you please provide examples of both working and faulty svg files?

Samz3n:
It is working when I save .svg files from Inkscape.

**PDF examples generated with mpdf:**
Because I changed:
`$this->svg_style = array( array( 'fill' => 'black', ... ) );` to `'fill' => 'yellow'`
You will see yellow shapes instead of black :)

**From Illustrator:**
// Different export settings (inline style, presentation attributes and tried to "Save As" and "Export As")
[1.pdf](https://github.com/mpdf/mpdf/files/1175992/1.pdf)
[2.pdf](https://github.com/mpdf/mpdf/files/1175993/2.pdf)
[3.pdf](https://github.com/mpdf/mpdf/files/1175990/3.pdf)
[4.pdf](https://github.com/mpdf/mpdf/files/1175991/4.pdf)

**From Inkscape:** 
// I needed to clean up the vector (outline text and release elements from compound path)
[5.pdf](https://github.com/mpdf/mpdf/files/1175994/5.pdf)


I also had some trouble with two different logos from Illustrator. 
I managed to get the colors right by using the inline-style option.

treedbk:
I had a similar problem where my company's logo was rendering in all black instead of the color specified in the svg file. This svg file was also exported from Illustrator. For some reason mpdf doesnt seem to recognize the color specified by the CSS in the svg file.

I fixed the problem by adding inline styles for each <g> tag. 

Each opening`<g>` tag needs to have an inline fill attribute: 

`<g style="fill: rgb(0,94,184);">` 

[...]

PR #1249, opened 2020-06-30
Fix issue #450, SVGs using a style tag, has styles ignored

antman3351:
Let's try again....


CHANGELOG.md
```diff
@@ -22,7 +22,7 @@ mPDF 8.0.x
 * Compound classes selector support, like `.one.two` or `div.message.special` (#538, @peterdevpl)
 * Fixed CMYK colors in text-shadow (#1115, @lexilya)
 * Skip non supported wrappers when resolving paths (#1204, @MarkVaughn)
-
+* Fixed SVGs using a style tag, has styles ignored ( Requires ext-dom ) (#450, @antman3351)
```
finwe, line 25: Please keep the empty line

src/Image/Svg.php
```diff
@@ -2824,6 +2824,93 @@ function svgWriteString($content)
...
+						if ($node->hasChildNodes()) {
+							$setStylesInline($node);
+						}
+					}
+				}
+			};
+
+			$setStylesInline($xml);
+
+		} catch (\Exception $e) {
```
finwe, line 2907: Where exactly can the exception be thrown? Can't you surround only relevant code portions to reduce indentation?

Also, catching \Exception directly is not recommended, DOM methods throw an \DOMException, so you should catch that only to prevent catching unrelated exceptions, especially when you throw the exception details away in the catch block.

tests/Issues/issue450Test.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+class Issue450Test extends \Mpdf\BaseMpdfTest
+{
+
+	/**
+	 * SVGs with global CSS styles, had the styles ignored
+	 */
+	public function testSvgWithGlobalStyles()
+	{
+		// Copied from  Mpdf\Image\MetricsGeneratorTest
```
finwe, line 27: How is this comment relevant?

tests/Issues/issue450Test.php
```diff
@@ -0,0 +1,63 @@
...
+		$otl = Mockery::mock(Otl::class);
+		$cssManager = Mockery::mock(CssManager::class);
+		$imageProcessor = Mockery::mock(ImageProcessor::class);
+		$this->sizeConverter = $this->createMock(SizeConverter::class); // Mockery::mock(ColorConverter::class); <-- doesn't work for me !??!
```
finwe, line 31: How are these comments relevant? Either find out why Mockery doesn't work and fix it, or do not comment at all.

finwe: Or state the reason of not working.

antman3351:
> Comments to the code are in the review. Also please fix the failing CI.

Fail CI is because of old (unsupported) version of PHPUnit isn't compatible with PHP 7.4. ( when calling $this->createMock( ... ); )

src/Image/Svg.php
antman3351: I thought DOMDocument was throwing an Exception, in fact it's issuing a Error (for malformed XML from another test) that PHPUnit was converting to an Exception.
I'll silence the Error as it's not much help in a PDF generation. `$xml->loadXML($data, LIBXML_NOERROR )`, the other option is to register an error handler to convert the Error to Exception. 

Surrounding the whole block I don't think is too bad as it's an "atomic" operation any Exception means something has gone wrong during the transformation. I prefer a SVG with missing colours to no SVG.  (Anyway I'll remove the try & catch)

I agree the indentation is terrible, I'll invert the ifs in the loop.

tests/Issues/issue450Test.php
antman3351: As I'm not familiar with the mpdf code base, if there's something wrong at least you know where I got it from 🤨

finwe:
Thanks for a push, merged in 9f67876

PR #1282, opened 2020-09-11
Fix typo in OtlDump.php

MultiSuperFreek:
On line 546, a variable $c was missing a $. This results in a undefined constant.


finwe:
Thanks!

PR #1289, opened 2020-09-25
fix: :bug: Don't throw exception, if backtrack_limit is disabled

dsturm:
If pcre.backtrack_limit is set to "-1" (disabled), the method `AdjustHTML` always throws an exception.


finwe:
Hi, thank you for the PR and sorry for the delay in reply.

I don't think disabling backtrack limit is a good thing. It will result in hard-to-explain delays in generating, or even endless loops.

I think it'd be better to alter the error message for the exception when the limit is `< 0` as done in 13d70e7.

Closing this.

Issue #1285, opened 2020-09-16
Undefined index level due to Directional Formatting charachter


stemis:
> Please use https://stackoverflow.com/questions/tagged/mpdf for all your general questions or troubleshooting!
> For contributing here, please see the guideline: https://github.com/mpdf/mpdf/blob/development/.github/CONTRIBUTING.md
>
> *Warning*: Failing to provide necessary information may cause the issue to be closed without consideration

### I found this bug 
We have been getting `Undefined index: level` errors from the following line:
https://github.com/mpdf/mpdf/blob/5dc65d646fe1fd59ba664da3fae070711639be01/src/Otl.php#L5634

What is causing this is the following unicode charachter: `\u{202C}` in combination with a few characters.

Not sure what the correct way is to deal with these charachters in a pdf context, but I do believe atleast a better error message can be put in place.

### This is mPDF and PHP version and environment (server/fpm/cli etc) I am using
PHP 7.4.7 (cli) (built: Jun 11 2020 18:33:07) ( NTS )
MPDF 8.0.7

### This is a PHP code snippet I use

```php
$html = base64_decode("PGh0bWw+CiAgIDxib2R5PgogICAgICAg2KfYqeKArNiMIAogICA8L2JvZHk+CjwvaHRtbD4=");
$mpdf = new Mpdf();
$mpdf->WriteHtml($html, 2);

```

sameeriqbal1200:
Hi all,
I have resolve this issue. 
Simply replace this **$bidiData[$i]['level']** with **$pel**
and this issue will resolved.
hope this will work.
Thank you

PR #1290, opened 2020-09-25
Undefined index level due to Directional Formatting charachter #1285

ramazantufekci:
Artık hata vermiyor.


finwe:
Thanks for the PR, please add a test case reproducing this change. see tests/Issues for reference.

PR #1291, opened 2020-09-25
keep_table_proportions should be false by default

janbaros:
According to documentation, default value for keep_table_proportions should be false. Otherwise, please, update documentation.

### DIFF
--- a/src/Config/ConfigVariables.php
+++ b/src/Config/ConfigVariables.php
@@ -266,7 +266,7 @@ class ConfigVariables
 			'ignore_table_widths' => false,
 			// If table width set > page width, force resizing but keep relative sizes
 			// Also forces respect of cell widths set by %
-			'keep_table_proportions' => true,
+			'keep_table_proportions' => false,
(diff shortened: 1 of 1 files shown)


finwe:
Where is this variable false in the documentation? https://mpdf.github.io/reference/mpdf-variables/keep-table-proportions.html. Feel free to reopen with additional information.

janbaros:
I've possibly misinterpreted statement in:

https://mpdf.github.io/tables/table-layout.html

Overriding the Auto- LayoutPermalink¶

Other methods to override the default layout algorithm.

$keep_table_proportions

If the table width is set greater than the page width allows, mPDF will by default ignore any defined sizes and attempt to auto-layout the table to the page width. This will result in relative column widths etc. being lost.

If you wish to maintain the relative proportions of the table, set $keep_table_proportions = true; this forces the table to be resized, but keep relative proportions. NB It also forces respect of cell widths set by %

finwe:
The wording can be confusing as the default value was changed along with mPDF 6.0. All other places in the documentation should be correct, though. May revisit the docs on the linked page, thanks.

Issue #58, opened 2015-11-05
Refactor configuration from class properties to some sort of ParameterBag class instance


finwe:

fulldecent:
Reference for ParameterBag interface: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/ParameterBag/ParameterBagInterface.php

Dasc3er:
I don't know if it may be useful, but a base for such a refactoring may be developed from the following structure (parsed from https://mpdf.github.io/reference/mpdf-variables/overview.html).

<details><summary>Click to show: Config.php</summary>
<p>

```php
<?php

namespace Mpdf;

class Config
{
	// Paging
	/**
	 * The alias $useOddEven was used before v4.0.
	 * Scope: DOCUMENT.
	 * @since 4.0
	 */
	protected $mirrorMargins = 0;

	/**
	 * Removed v6.0.
	 * @since 2.3
	 */
	protected $restoreBlockPagebreaks = false;

	/**
	 * Scope: DOCUMENT.
	 * @since 2.3
	 */
	protected $forcePortraitMargins = false;

	/**
	 * Scope: DOCUMENT.
	 * @since 2.3
	 */
	protected $displayDefaultOrientation = false;

	/**
	 * Scope: DOCUMENT.
	 * @since 3.1
	 */
	protected $autoPageBreak = true;

	/**
	 * Scope: DOCUMENT.
	 * @since 4.0
	 */
	protected $setAutoTopMargin = false;

	/**
	 * Scope: DOCUMENT.
	 * @since 4.0
	 */
	protected $setAutoBottomMargin = false;

	/**
	 * Scope: DOCUMENT.
	 * @since 4.0
	 */
	protected $autoMarginPadding = 2;

	/**
	 * Allows an (empty) end of block to extend beyond the bottom margin by this amount (mm).
	 * Scope: DOCUMENT.
	 * @since 5.5
	 */
	protected $margBuffer = 2;

	/**
	 * Adds date and page info for printer when using @page and marks:crop;.
	 * @since 5.1
	 */
	protected $printers_info = false;

	/**
	 * @since 5.1
	 */
	protected $bleedMargin = 5;

	/**
	 * Distance of cross mark from margin in mm.
	 * @since 5.1
	 */
	protected $crossMarkMargin = 5;

	/**
	 * Distance of crop mark from margin in mm.
	 * @since 5.1
	 */
	protected $cropMarkMargin = 8;

	/**
[...]

PR #1294, opened 2020-10-04
Changes inside of ConfigVariables to allow usage as config object

Dasc3er:
This PR changes the following structures:
 - Inside Config\ConfigVariables, the default array of the configuration is changed to properties of the class, added `get` and `set` methods
 - Added `$config` in Mpdf to allow direct use of the configuration object.

Related issue: #58
This is a simple start to separate the configuration from the Mpdf class.
Why the change? Using object properties can help with documentation (docblocks, type hinting, attributes, ...) and auto-completion features on different IDEs.
The change is quite small in relation to this PR, because there is not more information available on the features the Configuration object must have (quite vague issue).

What's next? If this PR is accepted, there can be a start in the conversion of the Mpdf properties usage to use directly the new `$config` object.

### DIFF
--- a/src/Config/ConfigVariables.php
+++ b/src/Config/ConfigVariables.php
@@ -12,516 +12,551 @@ use Mpdf\Ucdn;
 class ConfigVariables
 {
 
-	private $defaults;
+	// PAGING
+	protected $mirrorMargins = 0;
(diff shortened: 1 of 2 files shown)


Dasc3er:
Closed as unapproved to now and unmaintained.

PR #1297, opened 2020-10-09
Fixes bug on $str if provided key value is string and not int

alexmigf:


finwe:
Thanks for the PR. Please add a test case with code sample reproducing this fix (see `tests/Issues` for reference). Also, please use type casting `(int)` insteadof `intval` function. Then add a changelog entry.

finwe:

PR #1334, opened 2020-12-07
Validate UPCA barcode length

finwe:


arcanisgk:
Validate barcode length only will work with UPCA what happend with other standars??

finwe:
I don't understand your question.

arcanisgk:
you have add "Validate UPCA barcode length" I imagine this is an improvement for the UPC barcodes; what about other barcode standards like EAN etc ... will you add any Validation???

finwe:
You had a problem with UPCA. I made a fix for UPCA. If you see potential problems with other barcode types, you are most welcome to present a PR resolving these problems.

arcanisgk:
ok I'll help you; I will verify each of the standards that you support with mpdf; okay.

arcanisgk:
So this is still pending? like version 9 and 10 seem distant to me...

finwe:
Yes. And yes, they are.

PR #1366, opened 2021-01-29
Suppress warning on unlink cache file

janfejtek:
Since iterator iterates over all of the files in the directory, I believe there is a race condition when deleting cache file and warning is triggered.
This suppresses the warning

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -98,7 +98,7 @@ class Cache
 					&& $item->isFile()
 					&& !$this->isDotFile($item)
 					&& $this->isOld($item)) {
-				unlink($item->getPathname());
+				@unlink($item->getPathname());
 			}
 		}
 	}


finwe:
Did you encounter the race condition live or is it just an assumption (albeit likely correct)?

This will hide all other errors (eg. inability to remove file because of permissions). This solution is not good, closing.

janfejtek:
Yes it happens, that is why I created this pull request. I'm getting this error once or twice per day

With warnings like:
unlink(...vendor\mpdf\mpdf\src\Config/../../tmp\_tempSVG304475ffaafa47a6336.46448511_0.svg): No such file or directory
unlink(..vendor\mpdf\mpdf\src\Config/../../tmp\_tempSVG529885ffb31fe28c799.24500906_0.svg): No such file or directory
unlink(...vendor\mpdf\mpdf\src\Config/../../tmp\_tempSVG756185ffc0c479baa32.85582425_0.svg): No such file or directory

But when I look to the actual directory the files are not there so I think this is a race condition. Because most of the time it works without any error

Ok, I will open a issue

src/Cache.php
```diff
@@ -98,7 +98,7 @@ public function clearOld()
 					&& $item->isFile()
 					&& !$this->isDotFile($item)
 					&& $this->isOld($item)) {
-				unlink($item->getPathname());
+				@unlink($item->getPathname());
```
duchymor, line 101: `				@unlink($item->getPathname());
				if (file_exists($item->getPathname())) {
                    unlink($item->getPathname());
                }`

PR #1370, opened 2021-02-01
Added margin and footer to @page page-name :first

antman3351:
Added css properties odd-header-name, even-header-name, odd-footer-name, even-footer-name  to `@page page-name :first` selector

Created PdfContentHelper to help testing of created PDFs


antman3351:
Hello, just wondering if the PR is ok or not? 🤔🤔🤔

PR #1397, opened 2021-03-15
Add possibility to parameterize the size of the watermark

WalterLuis:


finwe:
Thanks for the PR, changes needed are summarized in labels ad build errors.

WalterLuis:
@finwe 
No no, thanks to you and sorry for the delay, I think that would be fine by now :D

src/Mpdf.php
```diff
@@ -10448,7 +10449,7 @@ function _dounderline($x, $y, $txt, $OTLdata = false, $textvar = 0)
 	/* -- WATERMARK -- */
 
 	// add a watermark
-	function watermark($texte, $angle = 45, $fontsize = 96, $alpha = 0.2)
```
finwe, line 10451: As internal methods miss visibility definition, changing `watermark` method signature by removing a parameter is a no-go or this could be merged at soonest in the next major release as a breaking change. Keep the method signature as it was.

src/Mpdf.php
```diff
@@ -10458,7 +10459,11 @@ function watermark($texte, $angle = 45, $fontsize = 96, $alpha = 0.2)
 			$this->watermark_font = $this->default_font;
 		}
 
-		$this->SetFont($this->watermark_font, "B", $fontsize, false); // Don't output
+		if (!$this->watermark_font_size) {
+			$this->watermark_font_size = $this->default_font_size;
```
finwe, line 10463: Changing a config setting internally is a shady technique an will lead to unexpected results. 

Use the method parameter (see below) and optionally validate its value.

src/Mpdf.php
```diff
@@ -13090,7 +13095,7 @@ function Footer()
 
 		/* -- WATERMARK -- */
 		if (($this->watermarkText) && ($this->showWatermarkText)) {
-			$this->watermark($this->watermarkText, $this->watermarkAngle, 120, $this->watermarkTextAlpha); // Watermark text
+			$this->watermark($this->watermarkText, $this->watermarkAngle, $this->watermarkTextAlpha); // Watermark text
```
finwe, line 13098: Use the value from mPDF settings when calling the method.

finwe (changes requested):
I've added a few more remarks on the code.

WalterLuis:
Thank you very much for your comments @finwe, I hope the changes are better

src/Mpdf.php
```diff
@@ -12970,12 +12977,18 @@ function SetFooter($Farray = [], $side = '')
 
 	/* -- WATERMARK -- */
 
-	function SetWatermarkText($txt = '', $alpha = -1)
+	function SetWatermarkText($txt = '', $alpha = -1, $size = 120, $pos = null)
```
finwe, line 12980: A nice attempt, but the same applies in this case too: this would be technically a BC break and applicable the soonest in 9.0.

src/Mpdf.php
```diff
@@ -12970,12 +12977,18 @@ function SetFooter($Farray = [], $side = '')
...
 	{
 		if ($alpha >= 0) {
 			$this->watermarkTextAlpha = $alpha;
 		}
 		$this->watermarkText = $txt;
+		$this->watermark_font_size = $size;
+
+		if (!is_array($pos)) {
```
finwe, line 12988: What about the possible 'F' value set in SetWatermarkText?

src/Mpdf.php
```diff
@@ -12970,12 +12977,18 @@ function SetFooter($Farray = [], $side = '')
...
+			$pos = [0, 0];
```
finwe, line 12989: Again, setting an arbitrary value to a config variable from a method call will lead to unexpected behaviour. I know, the SetWatermarkImage does the same, but that's not a reason to encourage this in a new code.

src/Mpdf.php
```diff
@@ -12970,12 +12977,18 @@ function SetFooter($Farray = [], $side = '')
...
 	/* -- WATERMARK -- */
 
-	function SetWatermarkText($txt = '', $alpha = -1)
+	function SetWatermarkText($txt = '', $alpha = -1, $size = 120, $pos = null)
 	{
 		if ($alpha >= 0) {
 			$this->watermarkTextAlpha = $alpha;
 		}
 		$this->watermarkText = $txt;
+		$this->watermark_font_size = $size;
```
finwe, line 12986: I really don't like overwriting the config setting. I understand that in the current state of code where everything is used as a global there is not much space to play with, but maybe a different solution could be devised?

finwe (changes requested):
A next round of change requests. For this to be merged before 9.0 which has no ETA, keep changes minimal.

finwe:

Issue #868, opened 2018-09-26
Refactor path handling


finwe:
Handling of local/remote paths should be more coherent and configurable. Random notes:

- local filesystem path vs URL part
- running online (from a base URL)/in console?
- blacklisting/whitelisting external resources
- openbasedir alternative

PR #1399, opened 2021-03-18
Added `fileContentLoader` configuration variable allowing to override default file_get_contents (Closes #672)

### LINKED ISSUE #672: file_get_contents not enabled on some hosters
My hoster disables file_get_contents() into php configuration. ("allow_url_open" directive setted to false).

This problem is current (see https://stackoverflow.com/search?q=allow_url_open+file_get_contents or )

I suggest to have replace file_get_contents() by a "custom"  function or a static method.
The code will check if file_get_contents is enabled, and if not, use another way to load contents.

In my case, I have replaced all  file_get_contents from MPDF classes with a custom function, and I use "curl" instead of "file_get_contents()".

Hope this helps

prollandoc:
As discussed in #672, this PR adds a configuration variable fileContentLoader that allows a user to override the default `file_get_contents` when gathering a file content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ mPDF 8.0.x
 * PHP 8.0 is supported since 8.0.10 (#1263)
 * Fix: First header of named page is added twice (@antman3351, #1320)
 * Added `curlExecutionTimeout` configuration variable allowing to `CURLOPT_TIMEOUT` when fetching remote content
+* Added `fileContentLoader` configuration variable allowing to override default file_get_contents
 
 mPDF 8.0.0
 ===========================
--- a/src/Config/ConfigVariables.php
+++ b/src/Config/ConfigVariables.php
@@ -517,6 +517,9 @@ class ConfigVariables
 			'curlProxyAuth' => null,
 			'curlUserAgent' => 'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:13.0) Gecko/20100101 Firefox/13.0.1',
 
+			'fileContentLoader' => function ($path) {
+				return @file_get_contents($path);
+			},
 			'exposeVersion' => true,
 		];
 	}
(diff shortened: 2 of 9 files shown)


finwe:
Thanks for the PR: Even though this is a kind-of a step in the right direction, it is not enough, the file content loader should have correctly defined interface and be checked against the interface.

This implementation misses the premise of the linked Issue #672 - the problem is not with `file_get_contents` per se, but with loading content over HTTP with `allow_url_fopen` disabled. There is no reason to just replace local allow_url_fopen calls with the loader.

Furthermore, there already is a `Mpdf\RemoteContentFetcher` class, just not used always when applicable.

This is related to https://github.com/mpdf/mpdf/issues/868 and will be resolved accordingly with the other issues hinted in the link. As this will result in potential BC breaks, it is targeted to the next major version, 9.0.

Given all the reasoning above, I am closing this.

prollandoc:
Thanks for the feedback @finwe, I agree that this should be properly interfaced.

For the rest, we are currently encountering a lot of warnings, due to the fact that remote `file_get_contents` end up with 400 errors (probably due to the lack of context `file_get_contents` sends in its requests, this StackOverflow issue describes it in detail https://stackoverflow.com/questions/43750021/php-file-get-contents-returns-with-a-400-error), and doesn't seem bypassable in the `ImageProcessor` code. 

I was hoping taking advantage of the issue I linked would resolve @m4ucoder's issue as long as mine.

Would you then have a recommendation regarding our issue? How to prevent `file_get_contents` from being called from `ImageProcessor`, or how to make it silent as it then fallbacks on `curl` and succeeds due to the high customization we can add? Would opening a new PR which would just make the `file_get_contents` silent at the moment be an option?

Anyway, thanks again.

Best regards

finwe:
Feel free to follow up in #672 with a response to my question: where exactly is file_get_contents exclusively used to get remote content? If you provide a working example, maybe we can alter the codebase to use existing content fetcher.

finwe:
(I am not a fan of silencing errors.)

prollandoc:
> (I am not a fan of silencing errors.)

Me neither, that's why I suggested another option :D 

Thanks, let's continue on the issue then :)

Issue #832, opened 2018-08-09
Can "GPLv2 only" change to "GPLv2 or any later version" instead ("GPLv3+" would be even better)


cliffordp:
(Continuing from https://github.com/mpdf/mpdf.github.io/issues/83#issuecomment-387563393)

Please consider replacing the "only" part with "or any later version"

https://www.gnu.org/licenses/identify-licenses-clearly.html is an excellent reference/reason why "only" is discouraged.

Thank you for considering.

finwe:
I'm afraid this would be rather difficult as we'd need to gain a consent from the original library author(s). I think we have tried a few years back and failed.

Maybe @jakejackson1 or @marclaporte remember more?

See also https://github.com/mpdf/mpdf/issues/15

cliffordp:
Didn't realize @finwe wasn't the original author. Thanks for the quick response, though.

Another reason why it might be difficult, even if you were, is if there were a lot of community contributions, I'm unsure if you'd have the authority to change the license to add "or any later version" or to fully move to v3+

Thanks for considering. Either way would be a great win for going forward!

marclaporte:
Dear @all,

I spoke to and exchanged emails with Ian and he accepted to change the domain name ownership (http://mpdf1.com/) and license of the documentation to something that is community-friendly. 

His condition was that we "continue to make available a stand-alone mPDF library, so that people can download a library which does not require installers etc. linked to a larger project." This is something we all want anyways so we are good. I didn't ask for a license change of his work on mPDF, but I believe he would agree in the same spirit as the documentation. His words: "Unlike the actual mPDF script, the documentation/manual is all my work. So I am happy to release to you on any licence you like - feel free to choose! "

@danielhjames suggested to me that we try to align the license of the code and the documentation. When I pondered on perhaps using CC BY 4.0 for the documentation, he wrote "That would be problematic when mixing code examples with documentation."
In another context, he wrote:
"That is kind of what the https://github.com/mpdf/mpdf-examples are. I'd suggest developing documentation from these, as they are GPL from releases under that license when the examples were part of the bundled source."

Why is a license change important? For Tiki Wiki CMS Groupware, we actually added an installer to get around licence issues (we are LGPL and can't bundle GPL) and size (mPDF is 50+ megs). While mPDF was the initial motivator for this, it's a useful and generic feature for all libs. But this was a lot of work and not all projects will do it.
https://doc.tiki.org/mPDF
https://doc.tiki.org/Packages

[...]

PR #1402, opened 2021-03-25
Use MIT License

MaximeMazet:


finwe:
Do you really think that changing the licence file is enough?

finwe:
For extensive discussion (and why this probably won't happen) see https://github.com/mpdf/mpdf/issues/832

Issue #1404, opened 2021-04-02
Retain class attribute of embedded svg element during conversion to img element


MarcelBolten:
# Feature request

It is possible to [use embedded SVG images](https://mpdf.github.io/what-else-can-i-do/images.html#svg-images). However, they get [exported to a file and are than loaded](https://github.com/mpdf/mpdf/blob/3b21745a7b0178c14bf512d79b790a4351c16bce/src/Mpdf.php#L26932-L26938) via an `<img>` element.

It is really great that this works but unfortunately all formatting from the style and/or class attributes are lost.
Of course the `<img>` element can be addressed by css to add some formatting but that will be applied to all images.

It would be great if the class attribute could be retained/transplanted from the `<svg>` to the `<img>` element. Maybe like so:
```php
preg_match_all("/(<svg.*?<\/svg>)/si", $html, $svgi);
if (count($svgi[0])) {
  for ($i = 0; $i < count($svgi[0]); $i++) {
    $file = $this->cache->write('/_tempSVG' . uniqid(random_int(1, 100000), true) . '_' . $i . '.svg', $svgi[0][$i]);
    preg_match("/class=['\"].*['\"]/iU", $svgi[0][$i], $class);
    $html = str_replace($svgi[0][$i], '<img src="' . $file . '"' . ($class[0] ? ' ' . $class[0] : '') . ' />', $html);
  }
}
```

If that is not desirable maybe a constant class could be added.
```php
$html = str_replace($svgi[0][$i], '<img src="' . $file . '" class="mpdf-embedded-svg" />', $html);
```

PR #1405, opened 2021-04-02
Retain class attribute of embedded svg element during conversion to img element

MarcelBolten:
See #1404


MarcelBolten:
@finwe is there anything else I can do to get this merged?
Guidance is very welcome.

src/Mpdf.php
```diff
@@ -26933,7 +26933,8 @@ function AdjustHTML($html, $tabSpaces = 8)
 		if (count($svgi[0])) {
 			for ($i = 0; $i < count($svgi[0]); $i++) {
 				$file = $this->cache->write('/_tempSVG' . uniqid(random_int(1, 100000), true) . '_' . $i . '.svg', $svgi[0][$i]);
-				$html = str_replace($svgi[0][$i], '<img src="' . $file . '" />', $html);
+				preg_match("/class=['\"].*['\"]/iU", $svgi[0][$i], $class);
```
finwe, line 26936: This will not work with single quote in double quoted class string or vice versa, a backreference of the first quote is needed. It also will not work with unquoted class string.

MarcelBolten: I changed the regex (please see https://regex101.com/r/UK6HvN/3 for more details about the regex).
It will now only match the class attribute of the svg element and it can handle
1. single or double quoted class string
2. single quote in double quoted class string and vice versa
3. escaped double quote in double quoted class string
4. escaped single quote in single quoted class string and
5. unquoted class string

However, 3, 4, and 5 did not work for me
```
[php:error] PHP Fatal error:  Uncaught DivisionByZeroError: Division by zero in mpdf/mpdf/src/Tag/Img.php:372
Stack trace:
#0 mpdf/mpdf/src/Tag.php(240): Mpdf\\Tag\\Img->open()
#1 mpdf/mpdf/src/Mpdf.php(13771): Mpdf\\Tag->OpenTag()
Mpdf\\Mpdf->WriteHTML()
```

due to an issue not related to this PR and if I get it right unquoted class strings don't seem to be anticipated if I look at e.g. [src/Image/Svg.php#L3131](https://github.com/mpdf/mpdf/blob/3b21745a7b0178c14bf512d79b790a4351c16bce/src/Image/Svg.php#L3131) but maybe (most likely) I miss something here.
```php
function ImageSVG($data)
{
...
preg_match_all('/(<(linearGradient|radialgradient)[^>]*)xlink:href\s*=\s*["\']#(.*?)["\'](.*?)\/>/si', $data, $links);
```

MarcelBolten:
Hi @finwe,
It has been some time since I made the requested changes. What do you think about the altered regex?
If the regex is to complicated a simpler approach could be to use a static class name like:
```php
$html = str_replace($svgi[0][$i], '<img src="' . $file . '" class="mpdf-embedded-svg" />', $html);
```

MarcelBolten:
Hi @finwe ,

Do you have some comment on this PR?

PR #1414, opened 2021-04-21
Fix TypeError in PHP8

PATROMO:
We get the error "Unsupported operand types: string & int" because convertPlain() return `string[]`.
Although the return value describes the following `@return bool|float[]`.
I think converting to a float should solve the problem.

### DIFF
--- a/src/Color/ColorConverter.php
+++ b/src/Color/ColorConverter.php
@@ -173,7 +173,7 @@ class ColorConverter
 		$c = false;
 
 		if (preg_match('/^[\d]+$/', $color)) {
-			$c = [static::MODE_GRAYSCALE, (float)$color]; // i.e. integer only
+			$c = [static::MODE_GRAYSCALE, (float) $color]; // i.e. integer only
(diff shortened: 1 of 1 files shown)


src/Color/ColorConverter.php
```diff
@@ -173,7 +173,7 @@ private function convertPlain($color, array &$PDFAXwarnings = [])
 		$c = false;
 
 		if (preg_match('/^[\d]+$/', $color)) {
-			$c = [static::MODE_GRAYSCALE, $color]; // i.e. integer only
+			$c = [static::MODE_GRAYSCALE, (float) $color]; // i.e. integer only
```
finwe, line 176: Given the regexp pattern (and the comment), float does not make sense.

PATROMO: @finwe You're right, but `return float[]` also makes no sense if a string is returned. All other methods also return a float. If it makes more sense for you to convert to `int` at this point, then I would also agree.

PATROMO:
I'll try to write a test for it.

PATROMO:
close because #1415

PR #1415, opened 2021-04-23
Fix TypeError (PHP8) and Warning (PHP7) by casting values in the methods of ColorConverter

BolZer:
This PR references PR #1414. The PR #1414 can be closed as this PR incorporates the changes in #1414.

In PHP8 a type error will be thrown if the array, which is passed to the pack-Method in ColorConverter:convert, contains something other than a float or int. This is behaviour is currently the case as the type hint in the doc block is not honored and the array may contain a string. 

This is the case for something like: 

`'rgba(var(--sk_highlight,18,100,163),1)'`

The convert method under PHP7 will throw the following warning: 

> Warning: A non-numeric value encountered 

In PHP8 the following error will be thrown:

>  Uncaught TypeError: Unsupported operand types

A simplified example of this behaviour can be found here:

https://3v4l.org/hZU29

The produced array for the aforementioned example by the method ColorConverter:convertPlain 
is this:

Screenshot 2021-04-22 at 12 01 47

In the linked example we see that the value gets interpreted as int(0)

I've added some test cases to reflect on the changes-.

However the changes doesn't adhere to the typehint in the doc block due to the fact that the color mode 
is included in the array and a casting of this value to float may introduce unexpected behaviour.

### DIFF
--- a/src/Color/ColorConverter.php
+++ b/src/Color/ColorConverter.php
@@ -206,12 +206,12 @@ class ColorConverter
 		$g = hexdec(substr($cor, 3, 2));
 		$b = hexdec(substr($cor, 5, 2));
 
-        return [
-            3,
-            (float) $r,
-            (float) $g,
-            (float) $b
-        ];
+		return [
+			3,
+			(float) $r,
+			(float) $g,
+			(float) $b
+		];
 	}
 
 	/**
--- a/src/Color/ColorConverter.php
+++ b/src/Color/ColorConverter.php
@@ -230,7 +230,7 @@ class ColorConverter
 				return [static::MODE_RGB, (float) $cores[0], (float) $cores[1], (float) $cores[2]];
 
 			case 'rgba':
-				return [static::MODE_RGBA, (float) $cores[0], (float) $cores[1], (float) $cores[2], $cores[3] * 100];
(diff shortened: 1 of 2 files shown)


tests/Mpdf/Color/ColorConverterTest.php
```diff
@@ -89,6 +89,8 @@ public function convertColorsProvider()
 			['inherit', false],
 			['transparent', false],
 
+            ['#CCC url("img_tree.gif")', "3\xcc\xcc\xcc\x00\x00"],
```
BolZer, line 92: Didn't see any test case which tests something like this. I saw it in one of the comments in the source and included it.

src/Color/ColorConverter.php
```diff
@@ -173,7 +173,7 @@ private function convertPlain($color, array &$PDFAXwarnings = [])
 		$c = false;
 
 		if (preg_match('/^[\d]+$/', $color)) {
-			$c = [static::MODE_GRAYSCALE, $color]; // i.e. integer only
+			$c = [static::MODE_GRAYSCALE, (float) $color]; // i.e. integer only
```
BolZer, line 176: This could be replaced by a cast to int, but a cast to int or float needs to happen.

PATROMO:
@finwe can you merge this?

PR #1417, opened 2021-04-26
Fix PHP 8 TypeError and PHP 7 Warning by casting values in the methods of ColorConverter

BolZer:
In PHP8 a type error will be thrown if the array, which is passed to the pack-Method in ColorConverter:convert, contains something other than a float or int. This is behaviour is currently the case as the type hint in the doc block is not honored and the array may contain a string. 

This is the case for something like: 

`'rgba(var(--sk_highlight,18,100,163),1)'`

The convert method under PHP7 will throw the following warning: 

> Warning: A non-numeric value encountered 

In PHP8 the following error will be thrown:

>  Uncaught TypeError: Unsupported operand types

A simplified example of this behaviour can be found here:

https://3v4l.org/hZU29

The produced array for the aforementioned example by the method ColorConverter:convertPlain 
is this:

Screenshot 2021-04-22 at 12 01 47

In the linked example we see that the value gets interpreted as int(0)

I've added some test cases to reflect on the changes-.

However the changes doesn't adhere to the typehint in the doc block due to the fact that the color mode 
is included in the array and a casting of this value to float may introduce unexpected behaviour.

### DIFF
--- a/src/Color/ColorConverter.php
+++ b/src/Color/ColorConverter.php
@@ -206,12 +206,12 @@ class ColorConverter
 		$g = hexdec(substr($cor, 3, 2));
 		$b = hexdec(substr($cor, 5, 2));
 
-        return [
-            3,
-            (float) $r,
-            (float) $g,
-            (float) $b
-        ];
+		return [
+			3,
+			(float) $r,
+			(float) $g,
+			(float) $b
+		];
 	}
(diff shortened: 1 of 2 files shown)


BolZer:
Sorry. Created the PR in the wrong Repository

PR #1418, opened 2021-04-27
Set the default language using `/Lang` entry

cuongmits:
Generated pdf file doesn't pass eAccessibility checking which should be fixed/updated:
Checker URL: http://checkers.eiii.eu/en/pdfcheck/
More detail: https://www.w3.org/TR/2012/NOTE-WCAG20-TECHS-20120103/PDF16#PDF16-examples
Screenshot from 2021-04-27 12-57-52

### DIFF
--- a/src/Writer/MetadataWriter.php
+++ b/src/Writer/MetadataWriter.php
@@ -332,6 +332,12 @@ class MetadataWriter implements \Psr\Log\LoggerAwareInterface
 		$this->writer->write('/Type /Catalog');
 		$this->writer->write('/Pages 1 0 R');
 
+        if (is_string($this->mpdf->currentLang)) {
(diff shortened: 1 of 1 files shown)


tests/Mpdf/Writer/MetadataWriterTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $this->metadataWriter = new MetadataWriter($this->mpdf, $this->writer, $this->form, $this->protection, $this->logger);
+    }
+
+    protected function tearDown()
+    {
+        parent::tearDown();
+
+        Mockery::close();
+    }
```
cuongmits, line 65: not sure what to test here

PR #1429, opened 2021-05-10
Fix a ksort() undefined type warning

### LINKED ISSUE #1428: OTL ksort expects array, null given
### I found this bug

When using the Manjari Google font with OTL features enabled, a PHP warning (fatal error in PHP8) occurs while trying to display the text `പ്ലാസ്ത്റില്`.

### This is mPDF and PHP version and environment (server/fpm/cli etc) I am using

`development` branch. 
PHP 7.4

Just needed an issue number for the PR, which contains the working example.

jakejackson1:
The fix is to predefine the $GlyphByClass. This was commented out in the code already, and included a dummy index. My thoughts are this was done haphazardly to try fix PHP notices in the past, before the codebase was somewhat modernised with appropriate isset() checks, and so I've opted for the empty array instead.

Resolves #1428


finwe:
Thanks, Jake!

Issue #1430, opened 2021-05-10
Quiet zones customization


vincenzo-mennella-shippypro:
# Feature request

Hi,
I need to created a mixed barcode made up of mixed `EAN128B` and `EAN128C`. 

Putting 
```html
<barcode code="1234" type="EAN128B" /><barcode code="5678" type="EAN128C" />
```
works but two quiet zones are added around each barcode which, of course, I don't need since it must look like a single barcode.

Is it possible to add a quiet zone customization functionality?

Thank you!

vincenzo-mennella-shippypro:
Hi! Do you have any feedback on this? Do you think this will be implemented? Any workaround to suggest?

Thank you!

finwe:
Note the help wanted tag. If you need this implemented, find a developer to propose a pull request.

vincenzo-mennella-shippypro:
> Note the help wanted tag. If you need this implemented, find a developer to propose a pull request.

Hi, I sent PR #1439 which implements this feature

PR #1439, opened 2021-05-19
Barcode quiet zones customization

vincenzo-mennella-shippypro:
Adds the functionality requested into #1430.

All existing tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ mPDF 8.0.x
 * Fix: First header of named page is added twice (@antman3351, #1320)
 * Added `curlExecutionTimeout` configuration variable allowing to `CURLOPT_TIMEOUT` when fetching remote content
 * Fix: Not all combinations were generated for more than three compound classes (@JeppeKnockaert)
+* Added `quiet_zone_left` and `quiet_zone_right` to barcodes which support quiet zones in order to customize its width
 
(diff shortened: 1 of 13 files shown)


finwe (changes requested):
This seems OK, just a few changes in addition to failing lint:

- Use null as a default/empty value, the empty string is quirky.
- Use nicer, more verbose and meaningful HTML parameter name and function parameter names and array keys. eg. `$quietZoneLeft`, `quiet_zone_left` and `QUIET_ZONE_LEFT`
- Add a test actually using the added code, with a code example that can be reproduced and perhaps even used in documentation

finwe:
Please, rebase instead of merging in upstream.

finwe (changes requested):
Great! Just one last bit, please use `=== null` comparison instead of `is_null`.

finwe (changes requested):
When fixing the lint, please squash commits to one, I'll merge right after.

vincenzo-mennella-shippypro:
Corrected the lint (sorry, didn't see it!)

finwe:
Thanks for quick responses.

vincenzo-mennella-shippypro:
Thank you for checking! Do you have an ETA for tagging this?

finwe:
I tend to create releases when a few features/fixes pile up, so no ETA.

PR #1441, opened 2021-05-20
Fix compound classes support when combining more than three classes

JeppeKnockaert:
When using more than three classes on an element, not all combinations were generated because of an issue with the combinations function in the Arrays helper class.

e.g. when generating the combinations for 'a', 'b', 'c' and 'd' (with k=2) the combination generating algorithm would correctly generate `ab`,` ac`,`ad` and then jump to `bd` and `cd` instead of generating `bc` as well, because it only increased the middle index, without resetting the last index first: `[0, 3]` became `[1, 3]` instead of `[1, 2]`.


finwe:
Thanks!

PR #1456, opened 2021-06-17
Update CssManager to use RemoteContentFetcher

greew:
The `CssManager` class didn't make use of the `RemoteContentFetcher` class, meaning that when it tried to fetch data via `curl`, it wouldn't all the options that the `RemoteContentFetcher` was able to use.

### DIFF
--- a/src/CssManager.php
+++ b/src/CssManager.php
@@ -52,7 +52,7 @@ class CssManager
 	 */
 	private $remoteContentFetcher;
 
-	public function __construct(Mpdf $mpdf, Cache $cache, SizeConverter $sizeConverter, ColorConverter $colorConverter, RemoteContentFetcher $remoteContentFetcher = null)
+	public function __construct(Mpdf $mpdf, Cache $cache, SizeConverter $sizeConverter, ColorConverter $colorConverter, RemoteContentFetcher $remoteContentFetcher)
 	{
 		$this->mpdf = $mpdf;
 		$this->cache = $cache;
@@ -2318,11 +2318,11 @@ class CssManager
 
 			$contents = @file_get_contents($localpath);
 
-		} elseif ($this->remoteContentFetcher !== null) { // if not use full URL
+		} else { // if not use full URL
 
 			try {
 				$contents = $this->remoteContentFetcher->getFileContentsByCurl($path);
-			} catch (MpdfException $e) {
+			} catch (\Mpdf\MpdfException $e) {
 				// Ignore error
 			}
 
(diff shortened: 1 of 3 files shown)


src/CssManager.php
```diff
@@ -2312,13 +2318,13 @@ private function getFileContents($path)
 
 			$contents = @file_get_contents($localpath);
 
-		} elseif (!$contents && !ini_get('allow_url_fopen') && function_exists('curl_init')) { // if not use full URL
+		} elseif ($this->remoteContentFetcher !== null) { // if not use full URL
```
finwe, line 2321: RemoteContentFetcher cannot be null when the constructor definition is changed. Is the ommited `!$contents` condition superfluous?

src/CssManager.php
```diff
@@ -47,7 +47,12 @@ class CssManager
 
 	var $cell_border_dominance_T;
 
-	public function __construct(Mpdf $mpdf, Cache $cache, SizeConverter $sizeConverter, ColorConverter $colorConverter)
+	/**
+	 * @var \Mpdf\RemoteContentFetcher
+	 */
+	private $remoteContentFetcher;
+
+	public function __construct(Mpdf $mpdf, Cache $cache, SizeConverter $sizeConverter, ColorConverter $colorConverter, RemoteContentFetcher $remoteContentFetcher = null)
```
finwe, line 55: RemoteContentFetcher should be mandatory.

CssManager is kind-of internal class so it is not a problem.

src/CssManager.php
```diff
@@ -2312,13 +2318,13 @@ private function getFileContents($path)
...
+		} elseif ($this->remoteContentFetcher !== null) { // if not use full URL
 
-			$ch = curl_init($path);
-			curl_setopt($ch, CURLOPT_HEADER, 0);
-			curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
-			$contents = curl_exec($ch);
-			curl_close($ch);
+			try {
+				$contents = $this->remoteContentFetcher->getFileContentsByCurl($path);
+			} catch (MpdfException $e) {
```
finwe, line 2325: Use a FQN on exception usage as a best practice.

finwe (changes requested):
This seems like a nice enhancement. Please follow up with comments.

src/CssManager.php
greew: Yep, the `!$contents` check is superfluous, as it has already been checked on line 2297.

Issue #1447, opened 2021-05-27
File handles not closed, possible memory leak (+ solution)


bilogic:
### I found this bug

Where file handles of imported PDFs are not closed after performing `->output()`, nor does there seem to be a way to close them.
This blocks me from deleting my source PDFs with the following exception `unlink(xGf23CbyXDm3HKy8baZ8p2iITjgxWPGZIN6SoM6m.pdf): Text file busy`

### This is mPDF and PHP version and environment (server/fpm/cli etc) I am using
mpdf/mpdf v8.0.11  
PHP 7.4.13 (cli) (built: Nov 24 2020 10:03:34) ( NTS )

### This is the PHP code snippet I use

I also found the solution here https://github.com/Setasign/FPDI/blob/cddf43dcfe57337bbdc687cb929021174e6696da/src/FpdiTrait.php#L90-L107

If I manually call this `->cleanUp()`, I can delete my PDF without exceptions.
However, there is another `->cleanup()` method in `Mpdf.php`
https://github.com/mpdf/mpdf/blob/58322c5cf83a5f86e0c5e2109af1e1ce47685b5d/src/Mpdf.php#L1549-L1553

So I'm not sure where is the best place to add and call this new `->cleanUp()` method.

bilogic:
@greew 

I would like to propose adding a `fpdiCleanUp()` below `src/Mpdf.php`'s `cleanup()`, would you accept this PR?
Thank you.

bilogic:
Sorry should have been @finwe thanks

finwe:
@bilogic I guess, I wasn't able to find a better solution. Please create a PR for docs (import & cleanup parts) too.

bilogic:
@finwe 

Ok great! 

Let me make the `fpdiCleanUp()` available as a first step, I'm not confident of touch `cleanUp()` yet due to a lack of understanding mpdf's internals. Perhaps `cleanUp()` can call `fpdiCleanUp()` or something.

Thank you.

finwe:
As you will have to duplicate fpdi cleanup functionality, why not just duplicate it in the mpdf cleanup() method directly?

bilogic:
Yea ok, I think I will do that.

bilogic:
@finwe done thanks!

finwe:
Closed by #1476

PR #1476, opened 2021-07-25
fixed file handle leak using Setssign's FpdiTrait::cleanUp()

bilogic:
See https://github.com/mpdf/mpdf/issues/1447

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -1550,6 +1550,14 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 	{
 		mb_internal_encoding($this->originalMbEnc);
 		@mb_regex_encoding($this->originalMbRegexEnc);
+
+		// this will free up the readers, based on code from Setasign's FpdiTrait::cleanUp()
+		foreach ($this->createdReaders as $id) {
+			$this->readers[$id]->getParser()->getStreamReader()->cleanUp();
+			unset($this->readers[$id]);
+		}
+
+		$this->createdReaders = [];
 	}
 
 	/**


src/Mpdf.php
```diff
@@ -1546,11 +1546,29 @@ public function __construct(array $config = [])
 		$this->selectoption = [];
 	}
 
-	public function cleanup()
-	{
-		mb_internal_encoding($this->originalMbEnc);
-		@mb_regex_encoding($this->originalMbRegexEnc);
-	}
+
+     /**
```
finwe, line 1550: Copying the comment as is does not make sense.

bilogic: sorry, don't you.. what should I do?

src/Mpdf.php
```diff
@@ -1546,10 +1546,19 @@ public function __construct(array $config = [])
 		$this->selectoption = [];
 	}
 
-	public function cleanup()
+	public function cleanup($allReaders = false)
```
finwe, line 1549: I'm not sure about the parameter. Let's assume, for this case, it to be always false so that the method signature is not changed.

So please, remove the parameter and change the code accordingly.

bilogic: ok, removed and pushed, hope the tests don't throw me curve ball again :)

src/Mpdf.php
```diff
@@ -1550,6 +1550,16 @@ public function cleanup()
 	{
 		mb_internal_encoding($this->originalMbEnc);
 		@mb_regex_encoding($this->originalMbRegexEnc);
+
+		// this will free up the readers, based on code from Setasign's FpdiTrait::cleanUp()
+		$allReaders = false;
+		$readers = $allReaders ? array_keys($this->readers) : $this->createdReaders;
```
finwe, line 1556: The allReaders variable is completely unnecessary, just use `array_keys($this->readers)` directly, please.

Also, please, squash all the commits and forcepush to the branch.

bilogic: Ok, sure. 

My thinking was since we or at least I'm not too sure what `$this->readers` or `createdReaders` really do, maintaining the entire snippet integrity can be helpful down the road.

I have not used squash or forcepush before, let me give it a try. Thank you for the patience.

bilogic: Ok, seems I managed both on first try. Since `$allReaders = false`, we should be looping over `$this->createdReaders`. Thanks.

finwe:
Thanks!

PR #1485, opened 2021-08-06
Added optional `continue2pages` parameter to `SetDocTemplate` method

bmg-ruudv:
Added optional `continue2pages` parameter to `SetDocTemplate` method, allowing a template to continue the last 2 pages alternately.

This option makes it possible to set a document template of (minimum) 2 pages, where the last 2 pages will continue to be used alternately as a template. This allows the creation of PDF's for booklets with a background where the left page of the booklet matches in the right page of the booklet.
For example, you can use a 2 paged PDF as a template, where page 1 is used for each left (odd) page of the booklet and page 2 is used for each right (even) page of the booklet.

I added a unit test, but as i am not really known with the PdfParser/PdfReader i based it on some of the other unit tests. So i hope it's ok, any feedback is appreciated. I created Letterhead3.pdf to use in the testscript, which is just a combination of Letterhead.pdf and Letterhead2.pdf.

Hope this functionality can be added to the main package so other users can use it too :) Feedback is appreciated.


bmg-ruudv:
I also added a pull request to update the documentation: https://github.com/mpdf/mpdf.github.io/pull/166

finwe:
Thanks for an exceptionally well prepared PR.

PR #1497, opened 2021-08-27
Ensure that all digits of a string are hexadecimal before decoding

derklaro:
This pull request resolves an issue with newer php versions printing a deprecation message (`Invalid characters passed for attempted conversion, these have been ignored`) when trying to decode invalid hex characters using the [hexdec](https://www.php.net/manual/function.hexdec.php) method.

The passed string is now validated to only contain hexadecimal characters before it gets passed to [hexdec](https://www.php.net/manual/function.hexdec.php) using [ctype_xdigit](https://www.php.net/manual/function.ctype-xdigit.php). If the passed string does not contain only hex chars the method returns `0` just as php does.


finwe:
Thanks!

PR #1505, opened 2021-09-15
Allow psr/log v2

snapshotpl:
v3 after merge this one.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
 		"ext-mbstring": "*",
 		"myclabs/deep-copy": "^1.7",
 		"paragonie/random_compat": "^1.4|^2.0|^9.99.99",
-		"psr/log": "^1.0",
+		"psr/log": "^1.0 || ^2.0",
 		"setasign/fpdi": "^2.1"
 	},
 	"require-dev": {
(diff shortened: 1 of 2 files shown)


finwe:
Do not add unrelated gitignore items.

finwe:
Any particular reason not to allow v3 at the same time?

finwe:
Also, please squash the commits to one.

snapshotpl:
v3 fails hard because:
```
Fatal error: Declaration of Mpdf\Mpdf::setLogger(Psr\Log\LoggerInterface $logger) must be compatible with Psr\Log\LoggerAwareInterface::setLogger(Psr\Log\LoggerInterface $logger): void in /usr/src/myapp/src/Mpdf.php on line 1569
```

Probably not impossible to support v3 before we not drop php under 7.1

finwe:
Ack. I'm OK with that. What is the reason for including the testlogger?

snapshotpl:
`RemoteContentTest` need test implementation

finwe:
Hmm... Can you create an implementation in a separate file under Mpdf namespace to remove the hassle with bracketed namespaces in _functions_ file and then use that implementation?

snapshotpl:
Not time for this perfection. However we can merge as it and separate if have time

finwe:
No time for merging this imperfection.

snapshotpl:
Too bad.

finwe:
Thanks for the current effort, anyway

PR #1513, opened 2021-09-23
Php8 support: Implementation of LoggerAwareInterface to match recent version

ynagtegaal:


finwe:
This cannot be done to retain lower PHP versions support.

PR #1564, opened 2021-11-18
Fix division by zero on some SVG

manusfreedom:
Corrects division by zero on some SVGs and if a point has a problem, this can be used to detect it.


finwe:
Thanks for the PR, please add a unit test reproducing this change. see tests/Issues for reference

finwe:

PR #1607, opened 2022-01-25
Suppress GD errors for PNG images

developerbmw:
Currently an ErrorException is thrown when trying to use an interlaced PNG and there's no way to prevent this from halting execution.

`imagecreatefromstring(): gd-png: libpng warning: Interlace handling should be turned on when using png_read_image`

This PR proposes to suppress the warnings from `imagecreatefromstring()` when loading PNG files. I notice the other file formats suppress the errors from this function, so it would be good to do this for PNG also for consistency.

### DIFF
--- a/src/Image/ImageProcessor.php
+++ b/src/Image/ImageProcessor.php
@@ -578,7 +578,7 @@ class ImageProcessor implements \Psr\Log\LoggerAwareInterface
 					return $this->imageError($file, $firsttime, sprintf('GD library with PNG support required for image (%s)', $errpng));
 				}
 
-				$im = imagecreatefromstring($data);
+				$im = @imagecreatefromstring($data);
 				if (!$im) {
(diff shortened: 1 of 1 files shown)


finwe:
Suppressing errors is never a good idea. Talk some sense to your error handler not to escalate this warning or to your PHP build not to throw this warning.

developerbmw:
Lol what? If you don't want to be suppressing errors then you should review the 18 existing error-suppressed function calls in that file. I was trying to bring some consistency to the table but never mind.

dragosprotung:
This error is because of this change in PHP itself https://github.com/php/php-src/pull/7615 and the recomended way of handling this is by silincing the call

developerbmw:
https://github.com/php/php-src/pull/8002

marnickmenting:
Resolved in [aabfaee](https://github.com/mpdf/mpdf/commit/aabfaeead028225c7bb4e4fb37d05c9b8a6f004c)

Issue #1589, opened 2021-12-31
Interlace handling should be turned on


VARGA-Peter:
Suddenly, I receive the below error message. What happened, how can I fix it?

`Warning (2): imagecreatefromstring(): gd-png: libpng warning: Interlace handling should be turned on when using png_read_image [ROOT/vendor/mpdf/mpdf/src/Image/ImageProcessor.php, line 552]`

Running under PHP 8.0.14

Klap-in:
Could it be that you used these interlaced pictures not before? so that the code does not change, but the stuff processed by the code?

Here a result from some googling:
https://mixable.blog/png-deactivate-interlace-to-avoid-libpng-warning-interlace-handling-should-be-turned-on-when-using-png_read_image/

(Please note, it is better to post troubleshooting questions in the Discussions section or at StackOverflow.)

VARGA-Peter:
> Could it be that you used these interlaced pictures not before? so that the code does not change, but the stuff processed by the code?
> 
> Here a result from some googling: https://mixable.blog/png-deactivate-interlace-to-avoid-libpng-warning-interlace-handling-should-be-turned-on-when-using-png_read_image/
> 
> (Please note, it is better to post troubleshooting questions in the Discussions section or at StackOverflow.)

Nope, nothing changed. Suddenly I receive the warning. I ran the code last time in September. Very strange.

VARGA-Peter:
I could localise the PNG image which causes this warning. It hasn't been changed since I am using mpdf and the last time in September there wasn't this warning. Can I post somewhere _privately_ the link to the image so you can check it?

VARGA-Peter:
Due to the fact the image didn't change since I use mpdf I suspect some change in the PHP settings which causes the warning to be shown. However, I fixed it using [ImageMagick](https://imagemagick.org/index.php) running this command found on [SO](https://stackoverflow.com/a/19742588/3179492):

`magick convert -interlace none interlaced.png uninterlaced.png`

dragosprotung:
This error is because of this change in PHP itself https://github.com/php/php-src/pull/7615

marnickmenting:
Resolved in [aabfaee](https://github.com/mpdf/mpdf/commit/aabfaeead028225c7bb4e4fb37d05c9b8a6f004c)

PR #1609, opened 2022-01-27
Missing silence operator

dragosprotung:
Silencing operator is normally used everywhere, but it's missing here

### DIFF
--- a/src/Image/ImageProcessor.php
+++ b/src/Image/ImageProcessor.php
@@ -578,7 +578,7 @@ class ImageProcessor implements \Psr\Log\LoggerAwareInterface
 					return $this->imageError($file, $firsttime, sprintf('GD library with PNG support required for image (%s)', $errpng));
 				}
 
-				$im = imagecreatefromstring($data);
+				$im = @imagecreatefromstring($data);
 				if (!$im) {
 					return $this->imageError($file, $firsttime, sprintf('Error creating GD image from PNG file (%s)', $errpng));
 				}


Klap-in:
It seems to be a recurring issue… But this solution is not yet satisfying everybody.
https://github.com/mpdf/mpdf/pull/1607
https://github.com/mpdf/mpdf/issues/1589

dragosprotung:
Everywhere calls to `imagecreatefromstring` are sinlenced but this one place. Is this call special in any way ?

Not silencing this call creates issues after https://github.com/php/php-src/pull/7615 was merged and the only solution for now it seems to be silencing it.

developerbmw:
I've submitted a PR to PHP to fix the interlacing warning that everyone is getting https://github.com/php/php-src/pull/8002

finwe:
Tried this out and the warnings discussed in mentioned pulls/issues are being output nontheless. Closing this, feel free to update with more solid implementation and/or reasoning

The correct solution for this is of course removing the silence operators in other places and solving potential errors more cleanly.

marnickmenting:
I understand a better error handling should be added, but I think this should be merged to be at least consistent with the other calls. There is no reasoning why this one is not supressed. Right now the whole output of PDF is aborted because one image is not loaded correctly, and there is no way to catch it without changing code within this library.

finwe:
Provide a reproducible example of this causing an error. I simply will not accept a PR just slamming an `@` at random places.

marnickmenting:
```
$mpdf = new \Mpdf\Mpdf();
$mpdf->AddPage();

// Will throw ErrorException: imagecreatefromstring(): gd-png: fatal libpng error: Read Error: truncated data
$mpdf->WriteHTML('<img src="https://attachment.freshdesk.com/inline/attachment?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpZCI6OTE1MzY2NTAyOSwiZG9tYWluIjoiYWN0djguZnJlc2hkZXNrLmNvbSIsImFjY291bnRfaWQiOjMwMzEzNX0.QT3sA6KgpJlOOR3RBziHZWrNzLc0O85ph-ajFyzuDU4">hallo');
```

Using this pull request, it outputs the PDF, but with a red cross as image, which is the best possible outcome in this case.

CleanShot 2022-04-05 at 14 09 13

By the way, I tried adding an @ before the WriteHTML statement and that works too. Of course it will supress even more, but might work as a work-around for @dragosprotung?

marnickmenting:
Putting a try/catch around it does not work, as it stops writing the HTML as soon as the error occurres, so data is missing. Placing an @ before WriteHTML or before imagecreatefromstring does not stop the writing of HTML.

finwe:
Resolved in aabfaee

marnickmenting:
> Resolved in [aabfaee](https://github.com/mpdf/mpdf/commit/aabfaeead028225c7bb4e4fb37d05c9b8a6f004c)

That's fast, awesome! Can I see somewhere when it will be tagged?

finwe:
in v8.1.0 which should be released in matter of days, I hope. https://packagist.org/packages/mpdf/mpdf https://github.com/mpdf/mpdf/releases

PR #1633, opened 2022-04-05
Prevent hyphenation of urls starting with https and E-Mail addresses

HKandulla:
Hi,
currently only URLs starting with `http:` or `www.` are prevented from hyphenation. I have added `https:` to the regex to include URLs starting with https.
Best, Hannes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ mPDF 8.0.x
 * Fix: Using mpdf in phar package leads to weird errors (#1504, @sandreas)
 * WEBP images support (#1525)
 * Add regex to prevent URLs starting with https from getting hyphenated
+* Prevent E-Mail addresses from getting hyphenated
 
 mPDF 8.0.0
 ===========================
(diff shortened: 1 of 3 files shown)


finwe:
Please, add unit tests and a changelog entry. Then rebase onto the development branch which fixes float errors in tests.

finwe:
Also, remove the `fix:` prefix from the commit message (as this is not entirely a fix) and start the message with a capital leter.

HKandulla:
Hi @finwe,
thanks for your quick response. I added unit tests and changelog entry and rebased onto development branch.
I also included [https://github.com/mpdf/mpdf/pull/1634](url) in this PR.
If there is anything else, let me know.
Best, Hannes

finwe:
Cool, thanks!

HKandulla:
Hi @finwe ,
I just realised that due to this part: [https://github.com/mpdf/mpdf/blob/cc1759f153c948c76b0ef54a39a6186be9965def/src/Mpdf.php#L8136](https://github.com/mpdf/mpdf/blob/cc1759f153c948c76b0ef54a39a6186be9965def/src/Mpdf.php#L8136)
URLs and E-Mail addresses still behave differently.

E-Mail addresses would break before Hyphen-minus.
URLs would not break at all

Do you think this behaviour is desired? I would suggest E-Mail addresses and urls should break before Hyphen-minus.
Please let me know whether I should prepare such a PR including tests, etc.
Thanks, Hannes

PR #1656, opened 2022-05-15
CSS error when running in localhost

minhnguyenerp:
Fixed "undefined index: host" exception when running in localhost


finwe:
This completely reverses the condition, makes no sense. Plus, missing example code in a test case.

PR #1657, opened 2022-05-16
Exception throw when running in localhost

minhnguyenerp:
Fixed "undefined index: host" exception when running in localhost

In my situation the $path parameter in the normalizePath function was "C:/Projects/xampp-win64-7.4.16-0-VC15/htdocs/ivy/web/css/fonts/open-sans/open-sans.css"
Then the $tr = parse_url($path); return Array ( [scheme] => C [path] => /Projects/xampp-win64-7.4.16-0-VC15/htdocs/ivy/web/css/fonts/open-sans/open-sans.css )
After that the exception "Undefined index: host" throw at the line if (!empty($tr['scheme']) && $tr['host'] && !empty($_SERVER['DOCUMENT_ROOT'])) {

### DIFF
--- a/src/CssManager.php
+++ b/src/CssManager.php
@@ -2304,7 +2304,7 @@ class CssManager
 
 			// WriteHTML parses all paths to full URLs; may be local file name
 			// DOCUMENT_ROOT is not returned on IIS
-			if (!empty($tr['scheme']) && empty($tr['host']) && !empty($_SERVER['DOCUMENT_ROOT'])) {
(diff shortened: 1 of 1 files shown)


finwe:
Still missing a test to reproduce.

minhnguyenerp:
[Here](https://www.youtube.com/watch?v=zMgUUeFtgJ8) is what I saw in my project, there is no test to reproduce.

finwe:
Unless knowing what exactly causes this, I will not merge a simple slapping of empty() into the code.

PR #1660, opened 2022-05-18
Fix "Undefined Array Key" Error for Multiple Columns

Michael-Labriola:
I periodically get an "Undefined array key -1" error when $bk = 0 and $bk-1 = -1.  I don't know if this is the best way to fix this, but changing the condition so $bk-1 is not referenced when $bk = 0 makes the error go away and the resulting PDF is still correct.

Thank you


Michael-Labriola:
What is the next step to get this going?

Klap-in:
As the label tells, the unit test is missing. Please make from your example a unit test. (see for examples e.g. https://github.com/mpdf/mpdf/tree/development/tests/Issues)

Michael-Labriola:
OK.  I created Issue1660Test.php.  Is that what I needed to do?

Issue #1606, opened 2022-01-24
Notice: Undefined variable: glyphIDtoUni


jruvinski:
### I found this bug
Notice: Undefined variable: glyphIDtoUni

### This is mPDF and PHP version and environment (server/fpm/cli etc) I am using
mPDF 8.0.17 ; PHP 7.4.27 ; Apache 2.4.52 (Laragon Local Server)

### This is the PHP code snippet I use
I am using with PhpSpreadsheet v1.21, in this example will save empty, in all cases with or without data is saving and showing the notice, code:

```
<?php

use PhpOffice\PhpSpreadsheet\IOFactory;
use PhpOffice\PhpSpreadsheet\Spreadsheet;

$spreadsheet = new Spreadsheet();
$writer = IOFactory::createWriter($spreadsheet, 'Mpdf');
$writer->save('aaa.pdf');

```

### This is the HTML/CSS code snippet I use

```
No html/css
```

This is the notice:

```
Notice: Undefined variable: glyphIDtoUni in C:\laragon\www\ppv\vendor\mpdf\mpdf\src\Mpdf.php on line 3990
```

I think is just declare the variable as null before the if on line 3916.

uestla:
I'm having the exact same problem - exporting PDF in browser works fine, but testing it in console va PHPUnit test fails on the same `Notice: Undefined variable: glyphIDtoUni` error.

After some digging I found out that the root cause is in `File::sysGetTempDir()` in PhpSpreadsheet library that calls `sys_get_temp_dir()` which returns different path when called via web server (i.e. in browser) and via console.

In console it returns my user directory which does not contain `*.gid.dat` font file which leads to the variable `$glyphIDtoUni` to be undefined:

https://github.com/mpdf/mpdf/blob/d3ef7689800c6356ab7df7604289f97c17f37c76/src/Mpdf.php#L3915-L3942

I fixed this by explicitly setting the `sys_temp_dir` configuration directive in `php.ini` to the same path as the web server uses.

But the issue of potentially undefined variable `$glyphIDtoUni` still remains.

vanjoechua:
if ($this->fontCache->has($fontkey . '.cw.dat')) {
				$cw = $this->fontCache->load($fontkey . '.cw.dat');
                $glyphIDtoUni = $this->fontCache->load($fontkey . '.cw.dat');
			}

mohdazrulali:
Clear cache pc.
Folder mpdf dalam C/users/..../appdata/temp/......
[...]

PR #1664, opened 2022-05-28
Fix for issue-1606

K-Mazur:
We're using "carlos-meneses/laravel-mpdf" wrapper over mpdf in our Laravel project. Some of us are getting ErrorException "Undefined variable: $glyphIDtoUni" error. Using a ternary operator here fixed the problem.

Related to #1606


Klap-in:
Could you please add an unit test as well?

finwe:
Closing as stale. Can reopen and assess when the test is added.

anasuryana (commented):
it works for me, but what is the root cause ?

Issue #1622, opened 2022-02-27
Support for psr/log 3


Magiczne:
# Feature request

Hi! Are there plans to bring psr/log 3 support for this library? Currently, cause of added return types in the `psr/log`  there is an error while using this lib.

```
Declaration of Mpdf\Mpdf::setLogger(Psr\Log\LoggerInterface $logger) must be compatible with 
Psr\Log\LoggerAwareInterface::setLogger(Psr\Log\LoggerInterface $logger): void
```

Since PHP 7.3 is no longer supported adding return types to the mPDF could be beneficial.

finwe:
Not at the moment, it doesn't work properly with older PHP versions which we plan to keep supporting (you cannot fulfill the contract of 3.x unless having PHP >= 8.0)

Magiczne:
Why bother supporting dead php versions? Supporting so old php versions only lead to more security problems out in the wild since people won't be eager to upgrade cause their favorite libs are supporting it.

finwe:
I am not going to participate this discussion again, mPDF will, for the time being, still support old PHP versions. There is plenty said about this in older closed issues.

szymekjanaczek:
While we're at cyber-sec aspect, lets take a look at this list of vulnerabilities in the earlier PHP versions:
[CVE Details](https://www.cvedetails.com/vulnerability-list/vendor_id-74/product_id-128/PHP-PHP.html)

Magiczne:
@finwe can you point me to this issues?

shooorov:
just update 'composer update'

aligomboz:
Hello my friend, did you solve the problem?

finwe:
Yes, this issue is closed.

aligomboz:
Can you tell me where the problem was because I am facing the same problem

finwe:
Issues do not serve for troubleshooting. Use discussions or ask at Stack Overflow.

finwe:
psr/log 3.0 is supported as of just released v8.1.6

PR #1668, opened 2022-06-15
Enable require package psr/log v3

onecentlin:
Since psr/log version is updated to version 3.0.0, enable to solve required dependencies.


finwe:
Did you even try to run this on supported versions? Or search for a similar issues?

https://github.com/mpdf/mpdf/issues/1622

Issue #834, opened 2018-08-13
I would like to add "auto" as a font-size option for <textarea>


mauricekindermann:
### I would like to add "auto" as a font-size option for <textarea>

- MPDF sets the font size for textareas at 12pt.  
- You can change this to 1pt+ / 1px+
- Setting font-size to 'auto' crahes the render, and '0' makes the textarea 0px by 0px
- PDFs support "auto" fontsize (0) which give you much more flexible/cleaner textareas
- Fun Fact: Saving a Word Doc as PDF does this by default, which is how I learnt about it

Here's a demo, mPDF left, "auto" font size right:

![auto_font](https://user-images.githubusercontent.com/3828998/44013264-e0e954e6-9f07-11e8-8bac-7b4432d80cd0.gif)

mauricekindermann:
That's finwe for tagging this properly, appreciate all your work!

I've been working with Textareas extensively, here's another textarea font-size related quirk I've found: 

1. Using rows="" + cols="" for Height + Width causes fonts to scale with the document, so the text inside the textareas remains somewhat the same, and the document expands or shrinks around it. For example: 

`<textarea rows='6' cols='15' style='font-size:30px;'></textarea>`

![2018-08-14_0941](https://user-images.githubusercontent.com/3828998/44064004-5847cc72-9fa6-11e8-998f-bd664ac01999.png)

2. However, using Width % + Height % stops this from happening, and gives you control over the actual font size within the textarea, for example: 

`<textarea style='font-size:30px; width: 14.28%; height: 8%;'>`

![2018-08-14_0944](https://user-images.githubusercontent.com/3828998/44064109-d06d2e5e-9fa6-11e8-8077-df746fef08c1.png)

finwe:
@mauricekindermann Please report this as a separate issue as it is out of scope of "auto" font-size.

mauricekindermann:
Done! Apologies.

QiUnknown:
I second this feature. Thanks for the library @finwe

finwe:
Resolved in #1721

PR #1721, opened 2022-09-16
Enable auto font-size for active forms

ChrisB9:
Solves issue https://github.com/mpdf/mpdf/issues/834

According to [this commit](https://github.com/foliojs/pdfkit/commit/131df9e0ae4bb7db8d3328a745a9acecdf74a311), it seems to be completely enough to just set the font-size parameter within the AcroForm definition to 0.

### DIFF
--- a/src/Tag/Input.php
+++ b/src/Tag/Input.php
@@ -70,7 +70,7 @@ class Input extends Tag
 		if (isset($properties['FONT-FAMILY'])) {
 			$this->mpdf->SetFont($properties['FONT-FAMILY'], $this->mpdf->FontStyle, 0, false);
 		}
-		if (isset($properties['FONT-SIZE'])) {
+		if (isset($properties['FONT-SIZE']) && $properties['FONT-SIZE'] !== 'auto') {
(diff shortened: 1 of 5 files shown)


finwe:
Thanks for the PR! Would you please add a changelog entry and a test with a code example showcasing the new functionality?

ChrisB9:
Done

src/Tag/Input.php
```diff
@@ -70,7 +70,7 @@ public function open($attr, &$ahtml, &$ihtml)
 		if (isset($properties['FONT-FAMILY'])) {
 			$this->mpdf->SetFont($properties['FONT-FAMILY'], $this->mpdf->FontStyle, 0, false);
 		}
-		if (isset($properties['FONT-SIZE'])) {
+		if (isset($properties['FONT-SIZE']) && $properties['FONT-SIZE'] !== 'auto') {
```
ChrisB9, line 73: I wonder, though, whether we should filter out `font-size: 0` too, as it breaks the input and text-area as well.

Maybe like this:
0 = default, auto = /F 0

finwe:
This works like a charm. Merging, thanks.

PR #1728, opened 2022-09-29
Tables colspan fix

frantisekmichlik:
PHP 8.1 html table colspan fix


finwe:
Please, fix CS, add a test case with a code example this PR fixes (see tests/Issues for reference). Changelog entry is not entirely necessary.

finwe:
Closing as stale. Can reopen and assess when the CI is fixed and the test is added.

PR #1735, opened 2022-10-06
Render empty textareas

### LINKED ISSUE #838: Empty Textareas Don't Render
1. If you don't an artificial blank space `&nbsp;` , empty textareas won't render. 

2. So, for now, I'm adding `&nbsp;` so that the textareas render and are usable. The issue is you then have to backspace each empty `&nbsp;` cell:

andreaswolf:
Resolves: #838

### DIFF
--- a/src/Tag/TextArea.php
+++ b/src/Tag/TextArea.php
@@ -149,6 +149,7 @@ class TextArea extends Tag
 
 		$this->mpdf->specialcontent = serialize($objattr);
 
+		$this->mpdf->ispre = true;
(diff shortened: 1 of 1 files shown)


finwe:
Thanks for the PR. Frankly, I am not entirely fond of changes to the code that work for some, but could cause problems for others. Especially when the method of resolving an issue is in using a seemingly unrelated internal variable. I'll try to test this, but I make no promises.

kevinpapst:
@finwe totally understand your reasoning, still would be nice if we could render an empty textarea somehow.

PR #1794, opened 2023-01-13
fix possible division by zero error

sadeq-n-yazdi:
Sometimes, when **mpdf** is going to find if it needs to embed the font file completely or embed substitute used characters of a font, it makes a `Division by Zero` error.
This is a fix for such a situation where there is some problem with getting character counts of documents or the characters count is zero.


finwe:
Why a 100? You have mistyped the name of the variable. A test is missing.

Issue #1817, opened 2023-02-07
Add option to disable autoScriptToLang on certain elements if wanted


Lkopo:
### Please describe the new functionality as best as you can.

Having `autoScriptToLang` globally is quite a fine function, but for some elements in the document, it might be a need to have only one language set for it so proper font is used and no language determination is done.

Example:

```html
<body>
  <table>
    <tr>
      <td lang="ja">ンターJA健康ー康</td>
    </tr>
  </table>
  <div>Hello 健康</div>
</body>
```

Currently, it's split into something like:
```html
<body>
  <table>
    <tr>
      <td lang="ja"><span lang="ja">ンタ</span>ーJA<span lang="und-Hans">健康ー康</span></td>
    </tr>
  </table>
  <div>Hello <span lang="und-Hans">健康</span></div>
</body>
```

Where a more desired output would be:

```html
<body>
  <table>
    <tr>
      <td lang="ja">ンターJA健康ー康</td>
    </tr>
  </table>
  <div>Hello <span lang="und-Hans">健康</span></div> <!-- process only this -->
</body>
```

I was thinking about some attribute on element level, e.g. `disableAutoScriptToLang` which will be taken into the account only if `autoScriptToLang` is set to `true` in mpdf configuration. It might work also on hierarchy level, so it means that this attribute will apply to all children elements too. In HTML template it might look like:

```html
<body>
  <table>
    <tr>
      <td lang="ja" disableAutoScriptToLang>ンターJA健康ー康</td>
    </tr>
  </table>
  <div>Hello 健康</div>
</body>
```

PR #1818, opened 2023-02-07
Add new attribute disableAutoScriptToLang #1817

Lkopo:
**Usage**:

```html
<body>
  <table>
    <tr>
      <td lang="ja" disableAutoScriptToLang>ンターJA健康ー康</td>
    </tr>
  </table>
  <div>Hello 健康</div>
</body>'
```

**Result**:

mPDF will not process content of `<td lang="ja" disableAutoScriptToLang>ンターJA健康ー康</td>` via `autoScriptToLang`. Also applies to inner elements:

```html
<body>
  <table>
    <tr>
      <td lang="ja" disableAutoScriptToLang>
        ンターJA健康ー康
        <span>健康</span> <!-- ignored too -->
      </td>
    </tr>
  </table>
  <div>Hello <span lang="und-Hans">健康</span></div> <!-- autoScriptToLang performed its action here -->
</body>'
```

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -26809,8 +26809,15 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 
 		$n = '';
 		$a = preg_split('/<(.*?)>/ms', $html, -1, PREG_SPLIT_DELIM_CAPTURE);
+		$disableAutoScriptToLang = false;
+		$elementLevel = 0;
+
 		foreach ($a as $i => $e) {
 			if ($i % 2 == 0) {
+				// ignore processing if autoScriptToLang was disabled on element level
+				if ($disableAutoScriptToLang) {
+					continue;
+				}
 
 				// ignore if in Textarea
 				if ($i > 0 && strtolower(substr($a[$i - 1], 1, 8)) == 'textarea') {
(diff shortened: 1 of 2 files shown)


jakejackson1:
@Lkopo thanks for making this contribution! Rather than a new attribute, would it make more sense to automatically disable `autoScriptToLang` if the element has a pre-defined `lang` attribute? 

For example:

```
<body>
  <table>
    <tr>
      <td lang="ja">ンターJA健康ー康</td> <!-- the language used for this element will always be `ja` -->
    </tr>
  </table>
  <div>Hello 健康</div>
</body>'
```

Lkopo:
@jakejackson1 I was thinking about it, but this might affect projects which use such setup and don’t want to have autoScript disabled. So I rather choosed a more safe option, but it’s not a problem to make it work that way too.

jakejackson1:
@Lkopo copy. 👍

It was my understanding the `autoScriptToLang` feature is designed to try to detect the correct language from the script. By explicitly defining a language with the `lang` attribute, someone is effectively saying "hey, anything inside this HTML tag is language X". Having `autoScriptToLang` override that implicit configuration could be considered a bug – which is precisely why you want to conditionally disable it.

Let's get @finwe to weigh in on the best path forward before any changes are made.

Lkopo:
I forgot to mention, that this is written in [docs for lang attribute](https://mpdf.github.io/fonts-languages/lang-6-x.html):

> The HTML lang attribute has a number of uses:
> 
> - when OTL tables are being used for a font, the language from the lang attribute is used to select which OTL features are applied;
> - used in conjunction with CSS :lang selector to allow CSS styles to be applied;
> - **can be used in conjunction with autoLangToFont and autoScriptToLang (cf.)**

So it looks like it’s supposed to work that way. But for me it also makes more sense that if I have language specified for some element, doesn’t make sense to run autoScriptToLang logic inside of it.

Issue #1826, opened 2023-02-20
Division by zero because of FontSizePt


sadeq-n-yazdi:
### Guidelines

- [X] I understand that [if I fail to provide all required details, this issue may be closed without review](https://github.com/mpdf/mpdf/blob/development/.github/CONTRIBUTING.md).

### Description of the bug

It seems some fonts (for Arabic characters) make the FontSizePt set to Zero. it is a font problem but it crashes the `mpdf` too. 
And because of that the following lines in the `Mpdf.php`, throw `Division by zero` fatal error:

```php
$charspacing = ($this->charspacing * 1000 / $this->FontSizePt);
$wordspacing = ($this->ws * 1000 / $this->FontSizePt) ;
```

this is part of the trace log for this error:
```text
 in Mpdf\Writer\FontWriter::writeFonts called at /var/www/html/vendor/mpdf/mpdf/src/Writer/ResourceWriter.php (113)
 in Mpdf\Writer\ResourceWriter::writeResources called at /var/www/html/vendor/mpdf/mpdf/src/Mpdf.php (9958)
 in Mpdf\Mpdf::_enddoc called at /var/www/html/vendor/mpdf/mpdf/src/Mpdf.php (2032)
 in Mpdf\Mpdf::Close called at /var/www/html/vendor/mpdf/mpdf/src/Mpdf.php (9449)
 in Mpdf\Mpdf::Output called at /var/www/html/php/src/Txxxxx/Controller/RxxxxxController.php (109)
 in Txxxxx\Controller\RxxxxxController::htmlToPdf called at /var/www/html/php/src/Txxxxx/Controller/RxxxxxController.php (42)
 in Txxxxx\Controller\RxxxxxController::generate called at /var/www/html/php/funcs.php (8195)
 in getUxxxxxSxxxxxRxxxxx called at /var/www/html/php/src/Txxxxx/Controller/MxAxxxxx/PxxxxxMxxxxxController.php (989)
 in Txxxxx\Controller\MxAxxxxx\PxxxxxMxxxxxController::getRxxxxx called at /var/www/html/php/src/Txxxxx/Controller/MyAccount/PxxxxxMxxxxxController.php (58)
 in Txxxxx\Controller\MxAxxxxx\PxxxxxMxxxxxController::anyIndex called at ? (?)
 in call_user_func_array called at /var/www/html/vendor/phroute/phroute/src/Phroute/Dispatcher.php (60)
[...]

PR #1827, opened 2023-02-20
#1826 - workaround for division by zero because of bad FontSizePt 

sadeq-n-yazdi:
Workaround for #1826 

This is a workaround for fixing division by zero error when the code detects font size is `0` (`mpdf::FontSizePt`) 

**Note**: According to [the document](https://mpdf.github.io/fonts-languages/default-font.html) the size `9` is the default font size.


finwe:
Add a test case displaying this faulty behaviour before the fix.
Furthermore, why not just prevent setting 0 to the FontSizePt property rather than arbitrarily checking its value on 6 places, and not even by a constant?

finwe:
Closing as stale. Can reopen when a test (or an example code) is added here or to the related issue.

Issue #1809, opened 2023-01-27
Support for psr/log 3.0


monta990:
### Please describe the new functionality as best as you can.

Support for psr/log 3.0

https://github.com/mpdf/mpdf/blob/ef90b4c104bb6b9c9f66884fa2afe8d62a18f429/composer.json#L31

Niklan:
+1 for that.

This block upgrading to Drupal 10 (https://www.drupal.org/project/pdf_using_mpdf/issues/3331918) which requires `psr/log:^3.0` because of Symfony 6. Even considering that module is used by at least 2000 website directly, and there are projects with direct dependency (without Drupal module) - it seems affects a lot of projects. Most likely, security support for Drupal 9 (Nov. 2023) will happen earlier than we receive mPDF 10.0 (#1637) and this put's projects at the danger. And more time passes, more projects will start facing this kind of issues.

If project tries to support that much verity of PHP versions, isn't it better to replace `psr/log` with own implementation that will work on all support PHP versions by this project? For now, like ~10 lines of code, where `psr/log`  interface is used, blocks whole library functionality and create issues like this. It's not worth it.

chesn0k:
I've created a [fork](https://github.com/chesn0k/mpdf/tree/psr-log-3) which works with psr/log:^3.0 if someone need it until this issue is resolved.

markdorison:
@chesn0k Is there any chance you could tag a release on the `psr-log-3` branch so it can be included via Composer without setting `minimum-stability: dev`? Having to set that flag could block downstream projects from upgrading to a version that requires the fork.

chesn0k:
@markdorison [I added a fork to packagist](https://packagist.org/packages/chesn0k/mpdf) so now you can just require `composer require chesn0k/mdpf`

Added tag `8.1.5`,  I will follow the releases and update the fork in the [development](https://github.com/chesn0k/mpdf) branch

markdorison:
Thanks @chesn0k!

finwe:
Thanks to @markdorison et al., psr/log 3.0 is supported as of just released v8.1.6

PR #1829, opened 2023-02-28
Support PSR Log v3

nbayramberdiyev:
This PR adds support for [`psr/log`](https://github.com/php-fig/log) v3.


finwe:
https://github.com/mpdf/mpdf/issues/1622

Niklan:
@nbayramberdiyev also, take a look at #1809 and provide your opinion as well, because #1622 is locked. This is definitely become a bigger problem the more time passes. There is also a fork with `psr/log:^3` support until mainline is bump it.

PR #1836, opened 2023-03-16
Remove " from Content-Disposition

develgold:
"Content-disposition" filename must be without " like lines: 27335, 27371


finwe:
> Additional directives are case-insensitive and have arguments that use quoted-string syntax after the '=' sign. Multiple parameters are separated by a semicolon (';').

PR #1848, opened 2023-04-01
Fix PHP 8.1 compatibilty issue

skills-up:
Check array key exists before comparing


skills-up:
This appears to be causing blank pages.

PR #1849, opened 2023-04-01
Fix PHP 8.1 compatibilty issue

skills-up:
Check array key existence before comparison


finwe:
Thanks for the PR, though it is not enough. Throwing an isset by itself might not be the best solution and in some cases might even break stuff.

Also, per contribution guidelines, kindly add a unit test with a simplified example HTML code showing the notice before and not showing it after the "fix" so that the error can be better understood and the fix better reviewed.

finwe:
Oh, and also squash the commits to one, the prior has no real meaning.

finwe:
Closing in favor of https://github.com/mpdf/mpdf/commit/865bca6d212486a7541270d27fc3f648805b12f4

Issue #1852, opened 2023-04-05
Upgrade support for PSR/LOG 3.0.0


valentinasbitcat:
### Please describe the new functionality as best as you can.

Psr/Log 3.0.0 is newest version, but mpdf/mpdf requires psr/log 1.0.0 or 2.0.0 and 3.0.0 is not supported.

Is there any time soon that this will be implemented?

![image](https://user-images.githubusercontent.com/129482368/229997516-7e5e2d01-95cf-48ba-9777-8b2b97b29e62.png)

finwe:
No. #1622

finwe:
Hint: you can also downgrade on your part to 2.0, unless some of your packages require 3.0 only.

finwe:
psr/log 3.0 is supported as of just released v8.1.6

PR #1853, opened 2023-04-05
Support for PSR/LOG 3.0

valentinasbitcat:
Solves this https://github.com/mpdf/mpdf/issues/1852 issue

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
 		"paragonie/random_compat": "^1.4|^2.0|^9.99.99",
 		"php-http/message-factory": "^1.0",
 		"psr/http-message": "^1.0",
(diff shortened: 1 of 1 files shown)


finwe:
Did you even try to install and run this on supported PHP versions? SMH. Not mentioning the trailing comma.

valentinasbitcat:
@finwe , yes this is what i get

image

finwe:
That's not what I meant. Changing supported version only is not nearly enough as the 3.0 introduces return types which have to be honored.

[This is how the pull request should have been done](https://github.com/mpdf/mpdf/pull/1829/files) if not for reasons mentioned in #1622

PR #1856, opened 2023-04-12
drupal 10 compatibility

o0Charly0o:
In drupal, the pdf Serialization module does not have a version for drupal 10, I am applying the necessary changes, for this we need to update this library, since in drupal 10, psr/log uses version ^3.0, I suppose there are other developments/modules that need this in drupal.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
 		"paragonie/random_compat": "^1.4|^2.0|^9.99.99",
 		"php-http/message-factory": "^1.0",
 		"psr/http-message": "^1.0",
-		"psr/log": "^1.0 || ^2.0",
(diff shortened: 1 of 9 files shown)


finwe:
Not possible, #1622

If Drupal insists on limiting all other libraries by supporting 3.0 only, too bad.

o0Charly0o:
Ok, I understand, sorry, I didn't know about that problem, just one question, do I understand that psr/log 2.0 is compatible with php8? I could comment on the drupal community to support psr/log 2.0 version and make mpdf compatible with drupal 10.

Thanks for your time

finwe:
Yes, psr/log 2.0 is completely compatible with PHP 8.

I doubt that Drupal will want to add 2.0 support, but you can certainly try.

Issue #1866, opened 2023-05-02
Variable to change color of watermark text


Ian-Robinson1:
### Please describe the new functionality as best as you can.

The watermark text color cannot be changed with a variable when a pdf is created. Currently, the only way to change it is to directly edit src/Mpdf.php line 10598, such as:

$this->SetTColor($this->colorConverter->convert(0, $this->PDFAXwarnings));

To:

$this->SetTColor($this->colorConverter->convert('red', $this->PDFAXwarnings));

Being able to change the color with a variable would be very helpful

v1har:
### 1. Need to define the variable:
_var $watermark_color;_


### 2. Set default in _construct:
_$this->watermark_color = 0;_


### 3. Create setter function
_function setWatermarkColor($watermark_color = 0){
$this->watermark_color = $watermark_color;
}_


### 4. Use variable in line 10484
_$this->SetTColor($this->colorConverter->convert($this->watermark_color, $this->PDFAXwarnings));_

v1har:
I am not a pro user of github so I thought it's easier to tell you the idea so you can implement / add pull request. I just tried to help because I need this functionality too. :)

PR #1872, opened 2023-05-25
Fix for issue: #1866

sdon2:
@finwe

But rather than stomping more and more public properties to the Mpdf class, why not introduce a method
`
public function setWatermark(Mpdf\Watermark $watermark): self
 `
with the Watermark object containing all details of the watermark? 

That's a refactoring consists lot of code changes and it needs lot of test to be passed. So, i've come up with this public property implementation that needs only a few lines of code change to fulfill the requirement.


finwe:
I'd rather not accept PRs that will make future refactoring harder just to save work now. Thanks for the effort, though

Saikishor164 (commented):
👍

PR #1888, opened 2023-06-12
Fixed Division by 0 issue in function applyGPOSpdf

peterkiama:


finwe:
What use case does this solve? Add a unit test reproducing the fixed behaviour.

PR #1890, opened 2023-06-13
Feature/stop warning on broken lists

ross-ritchey:
Hello,

I was working on a Magento 2 site that has MPDF integrated to make PDF's out of product listings.

A good number of the product descriptions on the site have LI's without a parent UL/OL (bad HTML - I know - but will take time to fix and can't guarantee a future user won't do this again)

When the PDF is going through MPDF, the BlockTag class has an issue where a WARNING level error is thrown in this instance. The warning is that an array key doesn't exist -> list_style_type (and presumably list_style_image and list_style_position as well - just errors before those)

This pull adds a check to see if any of those are not set and calls _setListMarker with default values ('disc' for list_style_type, the defaults noted higher in the file for the others).

This prevents the WARNING from being thrown - allowing a system like Magento 2 (which stops script execution when a WARNING is thrown) to properly finish generating the PDF.

Adding a pull request as I thought this fix might be helpful for others.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ New features
 Bugfixes
 --------
 
+* Fixed warning when LI is not wrapped in parent UL/OL
 * Better exception message about fonts with MarkGlyphSets (Fix for #1408)
 * Updated Garuda font with fixed "k" character (Fix for #1440)
 * Testing and suppressing PNG file conversion errors
--- a/src/Tag/BlockTag.php
+++ b/src/Tag/BlockTag.php
@@ -894,8 +894,15 @@ abstract class BlockTag extends Tag
 
 			$this->mpdf->listitem = [];
 
-			// Listitem-type
-			$this->mpdf->_setListMarker($currblk['list_style_type'], $currblk['list_style_image'], $currblk['list_style_position']);
+			// Sometimes LI's are created without the UL/OL tag around them. Bad HTML, but possible.
+			// With PHP 8.1 - this causes a WARNING to be thrown here, because the three array key's noted below do not get set up
+			// Some systems break execution with WARNING level errors - adding a check here to default this information
+			if (!isset($currblk['list_style_type']) || !isset($currblk['list_style_image']) || !isset($currblk['list_style_position'])) {
+				$this->mpdf->_setListMarker('disc', 'none', 'outside');
+			} else {
+				// Listitem-type
+				$this->mpdf->_setListMarker($currblk['list_style_type'], $currblk['list_style_image'], $currblk['list_style_position']);
+			}
 		}
 
 		// mPDF 6 Bidirectional formatting for block elements


finwe:
Hi. Thanks for the PR. But it obscures an underlying problem which could be somewhat seen without it. With it, the user will most likely never know something is wrong.

I'm hesitant to merge it without addressing this and adding an appropriate test suite. More generally, I'm always hesitant to merge just a bunch of isset checks around _things that cause a warning_.

I'd say a better course of action than to slap this band-aid on the warning issuing code without fixing it properly (either by non-wellformed HTML warnings or by initializing the list structures) would be to surround mpdf execution with a custom error handler catching and suppressing the warning.

ross-ritchey:
I agree with the sentiment here -> though my counterpoint is that there actually isn’t something wrong in these instances. Browsers and MPDF properly render list items without the surrounding UL/OL all the time. Yeah – that HTML it might not pass a validation check – but as far as the users are concerned it is working properly. Prior to PHP 8.1 this type of issue sent out a NOTICE, not a WARNING. IE – it was not an issue. PHP moved to warning’s for these to push developers to write clean code. This code is throwing a warning because it improperly assumes that the array keys will exist at that point -> IE – it is not written properly. I am open to the fact that my solution might not be the most ideal – however with this issue still in the code base, MPDF is not actually compatible with PHP 8.1.Thanks,Ross From: Matěj HumpálSent: Tuesday, June 13, 2023 2:30 PMTo: mpdf/mpdfCc: Ross Ritchey; AuthorSubject: Re: [mpdf/mpdf] Feature/stop warning on broken lists (PR #1890) Hi. Thanks for the PR. But it obscures an underlying problem which could be somewhat seen without it. With it, the user will most likely never know something is wrong.I'm hesitant to merge it without addressing this and adding an appropriate test suite. More generally, I'm always hesitant to merge just a bunch of isset checks around things that cause a warning.I'd say a better course of action than to slap this band-aid on the warning issuing code without fixing it properly (either by non-wellformed HTML warnings or by initializing the list structures) would be to surround mpdf execution with a custom error handler catching and suppressing the warning.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

finwe:
Notices and Warnings are the same in this context - neither of those stop the excecution of the code unless bumped up to an Exception by a 3rd party library, in your case Magento (I don't know about Magento, but most frameworks raise the Exception in development mode only, in production mode, the warning would be just logged and not shown and the PDF would be generated properly). So mPDF is really compatible with PHP 8.1, it's only the error_reporting level that is the factor here (and it eventually should not stop the execution of the code in production environments).

Still, this change:
- obscures information for the user that was present before 
    - call it a soft BC break, even, there were reported notices/warnings that helped users to fix their HTML
    - I'm musing if a logger call with the information about possibly non-wellformed HTML would suffice for me, I don't know yet
- is missing a test suite
- and most importantly IMHO could/should be solved by better data structure holding list information - with all properties initialized so that the isset calls are not necessary.
    - A need for this would be also obscured by this change and would be much harder to get back to

PR #1893, opened 2023-06-16
Fix PHP 8.x error

manusfreedom:
trborder-top and trborder-bottom are not always set.


finwe:
Thanks for the PR. Add a unit test failing before and passing after your change so that this can be reviewed properly. See `tests/Issues` directory in the repo. Also check and fix build failures.

Issue #1878, opened 2023-05-30
raise psr/http-message to version 2


pfpro:
### Please describe the new functionality as best as you can.

good evening. is is possible to raise "psr/http-message" in composer.json to version 2?

finwe:
It is, currently only the [same way the psr/log 3 support was done.](https://github.com/mpdf/mpdf/pull/1857)

Ph0tonic:
Hi,
I was willing to help push forward this feature when I realised all the work needed just to support PHP 5.6. I have read your previous discussions here https://github.com/mpdf/mpdf/issues/1263 (3 years ago), and the main argument was to support the wider range of PHP versions. By dropping php 5.6 only (end of maintenance 6 years ago) and keeping php ^7.0, developing this feature would be really simple as php 7 support return types. Here is the stats usage from Packagist, which I believe represents quite well stats of active development projects and php 5.6 represents less than 1%.
- https://packagist.org/php-statistics

From this perspective, I personally think that it might be time to move forward.

finwe:
If it were as easy as dropping support of PHP version, it would be done already. https://github.com/mpdf/mpdf/issues/1637

Ph0tonic:
Sorry, if I missed something but here is what I did when PHP 5.6 is dropped (see #1884).

finwe:
Thanks for your effort but not ATM.

NigelCunningham:
Starting work on a PR here: https://github.com/NigelCunningham/mpdf/tree/http_message_2_support

michaelw85:
It would be great if packages started following the [PHP release](https://www.php.net/supported-versions.php) cadence and stop encouraging using EOL versions.

For anyone who runs into fatal errors because other packages do install a modern PSR standard, here's a git patch for mpdf:

<details>
  <summary>PSR Patch</summary>

  ```
  diff --git a/mpdf/src/Http/Request.php b/mpdf/src/Http/Request.php
  index 6027daa..02f364d 100644
  --- a/mpdf/src/Http/Request.php
  +++ b/mpdf/src/Http/Request.php
  @@ -4,13 +4,15 @@ namespace Mpdf\Http;
   
   use Psr\Http\Message\StreamInterface;
   use Psr\Http\Message\UriInterface;
  +use Psr\Http\Message\RequestInterface;
  +use Psr\Http\Message\MessageInterface;
   
   /**
    * PSR-7 URI implementation ported from nyholm/psr7 and adapted for PHP 5.6
[...]

PR #1907, opened 2023-07-17
Support http-message 2.0 as well as 1.0.

NigelCunningham:
The 2.0 release only differs in adding return type hints and removing the strict types declarations. Either version should be usable with mpdf.

This update is also needed for projects that use pdf, where other dependencies already require http-message 2. We are seeking this need in the Drupal printable / pdf_api use of mpdf; the lack of http-message 2 support is preventing an update to Drupal 10.1 (https://www.drupal.org/project/printable/issues/3374990).


finwe:
See [this comment](https://github.com/mpdf/mpdf/issues/1878#issuecomment-1569636010), make this a draft and propose only when certain composer ci on all supported PHP versions passes.

NigelCunningham:
Yeah, that's what I started on - I initially thought the solution was going to be simpler since this one isn't overriding constructors; apologies for the premature PR. On the plus side, it hopefully provides visibility that something is happening and an opportunity for testers to jump in.

May I ask what might seem like a silly question from your perspective: why are you supporting such old versions of PHP?

finwe:
There are people still using them.

NigelCunningham:
Ok, and I can see there's no point going further with that question :)

finwe:
Maybe not even with this PR, over the last months, I've become much closer to releasing the next major version with older PHP versions dropped.

NigelCunningham:
Should I bother continuing then? I have plenty of other things I could spend the time on.

finwe:
Depends on if and how fast you need this, I'd give the drop a few months (which is quite fast given how the library is evolving)

NigelCunningham:
Ok. We have people wanting to use our Drupal module with D10 today, so I'll keep going.

finwe:
I'll create the repository for the module and give you access, so you can work under mpdf namespace directly.

NigelCunningham:
Oh, that would be great; thanks.

finwe:
@NigelCunningham Can you please mark this as a draft until there is actually anything to review? I'm getting notified for every commit.

NigelCunningham:
Ok, I've updated this PR and pushed to the http-message-aware-trait repo, having tested on Ubuntu with PHP 5.6, 7.0, 7.1, 7.2, 7.3, 7.4, 8.0, 8.1, 8.2.

It all works locally and I've also tested using the repository on github.

src/Http/Request.php
```diff
@@ -4,6 +4,7 @@
 
 use Psr\Http\Message\StreamInterface;
 use Psr\Http\Message\UriInterface;
+use mpdf\HttpMessageAwareTrait\HttpRequestTrait;
```
finwe, line 7: PascalCase, please. Also, order uses alphabetically.

composer.json
```diff
@@ -25,9 +25,10 @@
 		"ext-gd": "*",
 		"ext-mbstring": "*",
 		"mpdf/psr-log-aware-trait": "^2.0 || ^3.0",
+		"mpdf/http-message-aware-trait": "^2.0 || ^3.0",
```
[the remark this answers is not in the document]
NigelCunningham: Updated.

src/Http/Request.php
NigelCunningham: Resolved, thanks.

NigelCunningham:
I believe all issues have been resolved now. I've renamed the repo contents to mpdf/psr-http-message-shim. If you're able to rename the repo as well, everything should just work.

I've been testing with the following addition to the mpdf/mpdf composer.json:

diff --git a/composer.json b/composer.json
index 198b2c9..efc17ed 100644
--- a/composer.json
+++ b/composer.json
@@ -67,5 +67,11 @@
        },
        "config": {
                "sort-packages": true
+       },
+       "repositories": {
+               "mpdf/psr-http-message-shim": {
+                       "type": "vcs",
+                       "url": "https://github.com/mpdf/http-message-aware-trait.git"
+               }
        }
 }
 
 ... and have successfully checked tests pass with PHP 5.6, 7.0 through 7.4 inclusive and 8.0 and 8.1.

finwe (approved):

finwe:
Great job, thanks!

NigelCunningham:
You're welcome!

NigelCunningham:
Could I just query as to when you're planning another release? I can make the Drupal module that depends on this use the development branch for now, but it would be better to avoid that if I can :)

finwe:
Shortly. I just need to resolve one seemingly minor issue elsewhere to release 8.2.0

NigelCunningham:
Great; thanks!

finwe:
Released.

PR #1921, opened 2023-08-10
Fix footer links in PDF with different orientation pages

### LINKED ISSUE #1170: Links not clickable in footers with pages oriented differently
### I found this bug

Links in a footer are not consistently clickable in the generated PDF in major browsers (tested in Chrome, Firefox, Safari).

This issue appears to occur when there are pages with different orientations.

The links are clickable in a local PDF reader. (I'm using the Preview app on MacOS, but I assume Adobe Acrobat would work as well)

### This is mPDF and PHP version and environment (server/fpm/cli etc) I am using

`PHP version 7.3.12`
`mPDF version 8.0.4`

### This is a PHP code snippet I use

In this example, only one of the links is clickable in major browsers.

```
<?php

$mpdf = new Mpdf([
    'orientation' => 'L',
    'format' => 'Letter-L',
]);

$mpdf->WriteHTML('
<html>
<body>
<htmlpagefooter name="footer" style="display:none">
  <a href="https://google.com" target="_blank">Test Link</a>
</htmlpagefooter>
<sethtmlpagefooter name="footer" value="on" />
</body>
</html>
');

$mpdf->AddPage('P'); // Changing this to a landscape page makes both links clickable

$mpdf->WriteHTML('page 2');
```

AlbanRicherPro:
fix #1170

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -9748,6 +9748,13 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 				$this->margin_footer = $this->saveHTMLHeader[$n][$OE]['mf'];
 				$this->w = $this->saveHTMLHeader[$n][$OE]['pw'];
 				$this->h = $this->saveHTMLHeader[$n][$OE]['ph'];
+				if ($this->w > $this->h) {
+					$this->hPt = $this->fwPt;
+					$this->wPt = $this->fhPt;
+				} else {
+					$this->hPt = $this->fhPt;
+					$this->wPt = $this->fwPt;
+				}
 				$rotate = (isset($this->saveHTMLHeader[$n][$OE]['rotate']) ? $this->saveHTMLHeader[$n][$OE]['rotate'] : null);
 				$this->Reset();
 				$this->pageoutput[$n] = [];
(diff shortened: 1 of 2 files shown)


finwe:
Thanks for the PR. Can you elaborate on how this works exactly?

AlbanRicherPro:
Before the fix, when the PDF has a header/footer that contains a link and when the header/footer is displayed on both portrait and landscape pages in the PDF, the links are not properly displayed on landscape pages when the last page of the document is in portrait and the links are not properly displayed on portrait pages when the last page of the document is in landscape.


But all the headers/footers are only written in the PDF when calling the `Output` function that calls the `Close` function:
https://github.com/mpdf/mpdf/blob/938651fca46e9556324ba4d93bb106bd7b07380c/src/Mpdf.php#L9509

Then it calls the `_enddoc` function:
https://github.com/mpdf/mpdf/blob/938651fca46e9556324ba4d93bb106bd7b07380c/src/Mpdf.php#L2049

Then it calls the `_puthtmlheaders` function:
https://github.com/mpdf/mpdf/blob/938651fca46e9556324ba4d93bb106bd7b07380c/src/Mpdf.php#L9997

When it writes the headers/footers, the code loops on all pages without setting the orientation of the `current` page and that's why it keeps the orientation of the last written page:
https://github.com/mpdf/mpdf/blob/938651fca46e9556324ba4d93bb106bd7b07380c/src/Mpdf.php#L9718

Then it calls the `WriteHTML` function when writing the header:
https://github.com/mpdf/mpdf/blob/938651fca46e9556324ba4d93bb106bd7b07380c/src/Mpdf.php#L9767
and/or the footer:
https://github.com/mpdf/mpdf/blob/938651fca46e9556324ba4d93bb106bd7b07380c/src/Mpdf.php#L9847

This line in the `Link` function uses the `$hPt` class property:
https://github.com/mpdf/mpdf/blob/938651fca46e9556324ba4d93bb106bd7b07380c/src/Mpdf.php#L4326

And because the `Link` function is based on `$hPt` class property, the link in header/footer has not the correct `y` value calculated.

finwe:
Sorry for the delay and thanks for the great explanation. Seems solid and simple enough not to break much else, will merge and can revert if potential larger issues emerge.

Can you please just squash the commits to one and amend the message to end with `(Closes #1170)`?

Issue #1928, opened 2023-09-19
Embedded PNG broken due to incorrect chunk type detection


creatissimo:
### Guidelines

- [X] I understand that [if I fail to provide all required details, this issue may be closed without review](https://github.com/mpdf/mpdf/blob/development/.github/CONTRIBUTING.md).

### Description of the bug

When embedding a PNG image in the PDF, it is sometimes broken.

**Background**
I've investigated the issue and have found, that the issue is caused by the ImageProcessor -> processPng function.
In my case, the affected PNG has no gAMA chunk, but "gAMA" appears in one of the IDAT chunks.
The function uses strpos to check if a gAMA chunk exists, while it does take into account, that the gAMA chunk has to be placed before the first IDAT chunk according to the PNG spec (http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.Summary-of-standard-chunks)
In my scenario, this leads to a false positive gAMA hit and the incorrect $gamma value breaks the image.

**Solution**
Instead of using strpos on the whole image data for checking if a specific chunk type exists, the existing chunk types should be obtained by looping over all chunks and getting the type from the bytes 5-8.

### mPDF version

8.2.0

### PHP Version and environment (server type, cli provider etc., enclosing libraries and their respective versions)

PHP 8.2.10 (cli) (built: Sep  7 2023 05:57:10) (NTS)

### Reproducible PHP+CSS+HTML snippet suffering by the error

    $mpdf->WriteHTML('<html><head></head><body><img src="testimg.png" /></body></html>');
    $output = $this->mpdf->Output('', Destination::STRING_RETURN);
    file_put_contents("test.pdf", $output);

![testimg](https://github.com/mpdf/mpdf/assets/22854665/f9dc6aaf-62b2-4bc3-972e-8071f4391bd3)

PR #1929, opened 2023-09-19
Fix broken PNG due to incorrect chunk type detection #1928

creatissimo:
In order to fix https://github.com/mpdf/mpdf/issues/1928 I've refactored the ImageProcessor -> processPng function and replaced strpos function by getting chunk types from chunk bytes 5-8.
I've refactored all places and not only the gAMA section which was mentioned in the issue

### DIFF
--- a/src/Image/ImageProcessor.php
+++ b/src/Image/ImageProcessor.php
@@ -779,12 +779,32 @@ class ImageProcessor implements \Psr\Log\LoggerAwareInterface
 			$pngalpha = true;
 		}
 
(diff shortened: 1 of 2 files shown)


finwe:
Thanks for the PR! Other than the failing CS check, this looks good. Please fix the coding standard, squash the commits to one and end the commit message to `(Closes #1928)`, I'll then be happy to merge.

creatissimo:
Thx for your feedback. I've fixed the CS issues and squashed the commits with the commit message ending with `(Closes #1928)`. Please let me know if I can be of further assistance

PR #1933, opened 2023-10-13
Update Mpdf.php

jgudtmann:
Fixes background-size: cover when container is wider than image.


finwe:
Thanks for the PR. Though it lacks a test and more thorough explanation of what and how this should solve things. Please elaborate.

PR #1934, opened 2023-10-14
Fix Warning Undefined array key "VALUE" in Option.php

Shadow243:
Fixing an issue of the "Undefined array key "VALUE" notice in Option.php.
Link to the reported issue: https://app.glitchtip.com/evoludata/issues/2379545

### DIFF
--- a/src/Tag/Option.php
+++ b/src/Tag/Option.php
@@ -27,7 +27,8 @@ class Option extends Tag
 				$attr['VALUE'] = mb_convert_encoding($attr['VALUE'], $this->mpdf->mb_enc, 'UTF-8');
 			}
 		}
-		$this->mpdf->selectoption['currentVAL'] = $attr['VALUE'];
+        
+		$this->mpdf->selectoption['currentVAL'] = isset($attr['VALUE']) ? $attr['VALUE'] : $ahtml[$ihtml + 1];
 	}
 
 	public function close(&$ahtml, &$ihtml)
--- a/tests/Issues/Issue1934Test.php
+++ b/tests/Issues/Issue1934Test.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Issues;
+
+use Mpdf\Tag\Option;
+use Mpdf\Cache;
+use Mpdf\CssManager;
+use Mpdf\Form;
+use Mpdf\Otl;
+use Mpdf\TableOfContents;
+use Mpdf\SizeConverter;
+use Mpdf\Image\ImageProcessor;
+use Mpdf\Language\LanguageToFontInterface;
+use Mpdf\Color\ColorConverter;
+
+class Issue1934Test extends \Mpdf\BaseMpdfTest
+{
+	public function testWithFailingHtmlSnippet()
+	{
+		$html = '<select><option value="this option tag has the value">Option 1</option><option selected>Option 2</option></select>';
+		
+		$this->mpdf->WriteHTML($html);
+	}
+}


finwe:
Thanks for the PR. Though it lacks a test and more thorough explanation of what and how this should solve things. Please elaborate.

Shadow243:
Thanks for your feedback.

Regarding the issue with the provided link, it is currently protected by credentials, making it inaccessible for a complete understanding. To enhance clarity and provide a more comprehensive view of the problem, I have included a screenshot that offers additional context and detailed insights into the issue. This screenshot serves as a valuable supplement to the link and will aid in a more thorough comprehension of the matter at hand.
Screenshot 2023-10-18 at 23 50 06

finwe:
This does not provide any meaningful additional information at all.

Shadow243:
According to the function: `public function open($attr, &$ahtml, &$ihtml)
	{
		$this->mpdf->lastoptionaltag = '';
		$this->mpdf->selectoption['ACTIVE'] = true;
		$this->mpdf->selectoption['currentSEL'] = false;
		if (empty($this->mpdf->selectoption)) {
			$this->mpdf->selectoption['MAXWIDTH'] = '';
			$this->mpdf->selectoption['SELECTED'] = '';
		}
		if (isset($attr['SELECTED'])) {
			$this->mpdf->selectoption['SELECTED'] = '';
			$this->mpdf->selectoption['currentSEL'] = true;
		}
		if (isset($attr['VALUE'])) {
			$attr['VALUE'] = UtfString::strcode2utf($attr['VALUE']);
			$attr['VALUE'] = $this->mpdf->lesser_entity_decode($attr['VALUE']);
			if ($this->mpdf->onlyCoreFonts) {
				$attr['VALUE'] = mb_convert_encoding($attr['VALUE'], $this->mpdf->mb_enc, 'UTF-8');
			}
		}
		$this->mpdf->selectoption['currentVAL'] = $attr['VALUE'];
	}`

There is a condition checking if $attr['VALUE'] is set but while assigning $attr['VALUE'] to $this->mpdf->selectoption['currentVAL'] there is no.

We could change it like this: $this->mpdf->selectoption['currentVAL'] = $attr['VALUE'] ?? null; instead of adding it inside the condition. The issue here is that we are trying to access an undefined array key VALUE

finwe:
Cool, now add a test to the PR with a code example that is affected by this fix. See test/Issues for examples.

Shadow243:
Missing Test added.

finwe:
This way of testing works, I guess, but I still would like a test to present an actual failing HTML snippet.

Shadow243:
Test with failing HTML snippet added too. 

The HTML code 
```
<select>
 <option value="this is valid">Option 1</option>
 <option selected>Option 2</option>
</select>
```
contains an <option> element without the 'value' attribute defined, and this specific structure is the direct cause of the error. 

This test should pass without applied correction.

```
public function testWithFailingHtmlSnippet()
{
   $mpdf = $this->mpdf;

   $html = '<select><option value="this is valid">Option 1</option><option selected>Option 2</option>. </select>';

   try {
	$mpdf->WriteHTML($html);
   } catch (\Throwable $th) {
	$this->assertEquals('Undefined array key "VALUE"', $th->getMessage());
   }
}
```

Shadow243:
I changed also this: 
```
$this->mpdf->selectoption['currentVAL'] = $attr['VALUE'] ?? null; 
```
to 
```
$this->mpdf->selectoption['currentVAL'] = isset($attr['VALUE']) ? $attr['VALUE'] : null;
```
to support php version 5.x

finwe:
Great. Now that we know what HTML causes this, let's make this even more up to standard:

> [value](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option#value)
> The content of this attribute represents the value to be submitted with the form, should this option be selected. If this attribute is omitted, the value is taken from the text content of the option element.

So, when the value attribute is not set, the textual value of the element should be used. Can we easilly set that instead of a default null?

Shadow243:
Requested updates Done.

With this, we are putting the textual value instead of null:
```
$this->mpdf->selectoption['currentVAL'] = isset($attr['VALUE']) ? $attr['VALUE'] : $ahtml[$ihtml + 1];
```

Shadow243:
@finwe I'm waiting for your review to make sure every thing is now okay.

PR #1940, opened 2023-10-25
fix: FontWriter PDF/A/X Font Subsetting should be enabled (Closes #1315)

### LINKED ISSUE #1315: font subsetting in PDF/A PDF/X
Is there a specifc reason for this?

https://github.com/mpdf/mpdf/blob/822c72c1e5811827f0010f556a7c613f7904dd84/src/Writer/FontWriter.php#L65-L67

If not I would like to submit a pull request that introduces a config variable for having Font subsets for PDF/A or PDF/X files

[PDF/A - Page 6](https://www.pdfa.org/wp-content/uploads/2013/05/PDFA_in_a_Nutshell_211.pdf)
```
PDF/A also places higher demands on
the information it contains. All required
fonts (or at least all glyphs for the specific
characters used) must be embedded within
the PDF.
```

[PDF/X - Page 6](https://www.pdfa.org/wp-content/uploads/2017/05/PDFX-in-a-Nutshell.pdf)
```
Fonts must be embedded (either fully
embedded, or as an embedded subset
in which all characters used in the text
are present).
```

danielrona:
closes #1315


finwe:
Thanks for the PR. Why not just remove the if completely and retain other checks?

finwe:
Also, please:

- Add ` (Closes #1315)` to the commit message
- Add a changelog entry

NicolasCARPi:
Yeah, just remove the three lines!

NicolasCARPi:
@danielrona Are you going to address our comment or should we close this and make a new PR?

danielrona:
Sorry, read and forgot about this.

Removing those three lines would not make the subsetting happen, by default [only very large Fonts](https://github.com/mpdf/mpdf/blob/d748f1e90eb98a3f61d8a5ae8035ba1972ef268c/src/Writer/FontWriter.php#L57-L62) are being subset.

If you only use Arial Unicode MS you would never Subset the Font.

NicolasCARPi:
Does it make sense to **always** subset the font, irregardless of the ttf size? (maybe with a setting?)

danielrona:
> Does it make sense to **always** subset the font, irregardless of the ttf size? (maybe with a setting?)

I guess that would work too, but people might start complaining then about large PDF files when not using PDF/A or PDF/X

NicolasCARPi:
> people might start complaining then about large PDF files when not using PDF/A or PDF/X

I don't understand your reply. I'm suggesting always subsetting the font. For all kind of pdfs. Another way to ask that question would be: what is the downside of subsetting the font? Because the way I see it, it should always be done.

danielrona:
Sorry, never mind the comment, my bad - I guess it's time to head to bed...

PR #1980, opened 2024-01-11
Add `total page number` to prevent display Watermark Image on footer [ personal used only ]

theavuth9:
### My purpose of this change is to remove Image Watermark from the footer of all document


finwe:
This is an adhoc change that overall does not make sense to be incorporated to the library. Won't merge.

PR #2013, opened 2024-05-24
fix "page-break: avoid;" in tables when the avoid blocks are bigger t…

carli2:
…han the page size

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -22244,6 +22244,11 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 						// pagebreak-after is mapped to pagebreak-before on i+1 in Tags/Tr.php
 						$pagebreaklookahead++;
 					}
+					// corner case: if the pagelookahead is bigger than the pagesize, we break anyway, so fill up the page
+					if ($pagebreaklookahead * $maxrowheight + $extra > $pagetrigger + 0.001) {
+						$pagebreaklookahead = 1;
+					}
+					// if we exceed page boundaries: restart table on next page before printing the line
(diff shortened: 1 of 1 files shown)


finwe:
Thanks for the PR. As per contribution guidelines, add a test running a minimal example HTML code affected by this fix.

carli2:
I added a unit test that demonstrates what goes wrong without the fix

carli2:
This is a new pull request with kinda mixed commits. Important is the support for "%%" in SizeConverter since MS Outlook creates faulty CSS which we should tolerate.

finwe:
If you have separate issues, create separate pull requests, each with its own tests.

To warn you, I will be hesitant to incorporate quickhacks for invalid input that can be solved on user side.

carli2:
I removed the latest commit from this pull request. Test case is present. Awaiting merge.

carli2:
What is the current status of the review process?

finwe:
On hold. Will need to find time to consider how this might impact non-corner-case uses, as it seems hacky.

carli2:
Maybe I should give some more details what the commit actually does.

sometimes, the user provides big page-break:avoid blocks that do not fit a single page. The old algorithm was faulty because it just printed the first row of each "avoid" block, the second row on the next page and so on.

With the fix, the mpdf detects that the block cannot avoid a break and thus splits it up anyway.

I also added a test case via https://github.com/mpdf/mpdf/pull/2013/commits/d60c35906a7a03b8559071cc4c72ae3aefa9f81d which shows that the fix drastically reduces the amount of pages spilled with blocks being too big.

A general flaw of the current page break system is that mpdf takes the biggest row (`$maxrowheight`) to estimate the amount of rows to print. An exact break would be better especially in cases where rows have varying height (like in invoices with a lot of detail text in one of the rows)

carli2:
When will this pull request be merged? It is very important to not waste paper on many printouts.

carli2:
This is a reminder. Last time, I implemented the page-break: avoid; behaviour in tables for mpdf (which got merged, that's fine) but I found a bug which i fix with this pull request:

When the page-break: avoid; block is too large for one page, mpdf used to spawn one per page, which is considered unwanted. This patch fixes this behaviour by detecting large page-break blocks and forcing a break at exactly the page limit.

**please merge**, then I can close this fork and switch all my projects back to mainline mpdf.

finwe:
@carli2 Please cleanup the commit history from the merges and reverts (preferably by squashing to a single commit) and rebase on the main development branch, I'll then review

carli2:
thanks a lot!

PR #2024, opened 2024-06-13
Fix undefined offset in printobjectbuffer method

pmenotti:
This pull request addresses an "Undefined offset: 1" error that occurs in the printobjectbuffer method of the Mpdf class. The error happens when accessing an array without verifying if the index exists, which leads to an "Undefined offset" notice.

Changes Made:

Added a check to ensure the index exists before accessing it in the printobjectbuffer method.
File Modified:

src/Mpdf.php
Specific Changes:

At line 7289, added a check to verify if the array index 1 exists before accessing it:
php
Copy code
if (isset($vv[1])) {
    $translate_y = $this->sizeConverter->convert($vv[1], $maxsize_y, false, false);
    $tr2 .= $this->transformTranslate(0, $translate_y, true) . ' ';
}
Reason for Changes:
The error was causing issues during the execution of the printobjectbuffer method when the array did not have the expected number of elements. By adding this check, we prevent the "Undefined offset" notice and ensure the method runs smoothly.

Testing:
Tested the changes with various inputs to ensure no "Undefined offset" notices are thrown and the function behaves as expected.

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -7264,65 +7264,65 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 				// mPDF 5.7.3 TRANSFORMS
 				$tr2 = '';
 				if (isset($objattr['transform'])) {
-					$maxsize_x = $w;
-					$maxsize_y = $h;
-					$cx = $x + $w / 2;
-					$cy = $y + $h / 2;
-					preg_match_all('/(translatex|translatey|translate|scalex|scaley|scale|rotate|skewX|skewY|skew)\((.*?)\)/is', $objattr['transform'], $m);
-					if (count($m[0])) {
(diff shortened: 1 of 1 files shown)


finwe:
Thanks for the PR.

- Fix CS errors
- Comments in english or none at all
- Add a test case with a code this is supposed to fix (nb failing before the fix and failing after. see `tests/Issues` for inspiration)

finwe:
but looking at the source code and translateY docs, translateY should only have one parameter so this change will just hide the error and not make it work. If I understand it correctly, the $vv[1] used should be changed to $vv[0] instead.

But the whole code is so low quality that a better solution altogether would be best.

finwe:
Fixed in 5f861dc

PR #2043, opened 2024-08-15
Fix error message property in Mpdf.php

cebe:
Exception thrown on invalid UTF-8 input asks about property `in_charset`. The actual property name is `charset_in`:

https://github.com/cebe/mpdf/blob/dfb5346e2e7ffdb4b0f6d588e94629ad52fb3466/src/Mpdf.php#L529


finwe:
Thanks for a PR. Please add a unit test covering this change.

chezinut:
@finwe it looks like that is fix of mistype... and it stuck for ~2years

finwe:
@chezinut It apparently was not critical enough for anybody to be fixed according to guidelines ~2years back. If it is for you, feel free to propose a PR with the test present.

When the configuration property names change, this is a candidate to be overlooked again, so, with a test, that could be prevented.

PR #2045, opened 2024-08-21
refactor: clean code and improvements

ddevsr:


finwe:
Would you mind at least running tests locally before pushing?

PR #2046, opened 2024-08-28
Some of the Myanmar Fonts can't read Mpdf 

Khant-Nyar:
I am using pyidaungsu-2.5.3.ttf for myanmar fonts 
in database have some unreadable char so i little changes in code it worked !


finwe:
Please add a unit test with a minimal example code mitigated with this fix. Looking at this, it seems too random and potentialy breaking something for other users.

PR #2051, opened 2024-09-12
Be more graceful with faulty HTML (created from famous email clients) which are interpreted by all browsers gracefully, too

carli2:
A famous email client creates quotations with `<blockquote style="width: 100%%">`. This is faulty CSS however all browsers I tested interpreted %% as % but not as px. Mpdf however defaults to "px" interpretation which leads to 100px wide blockquotes which in consequence creates PDF documents with a very small stripe of text, lots of line breaks and a large number of pages.

So we should mimic the browser's behaviour and interpret CSS measurements with %% as unit as percent rather than pixels.


carli2:
This pull request also includes a unit test

carli2:
okay, I will reopen based on mpdf/development

finwe:
You could have rebased and force-pushed, but whatever works.

PR #2055, opened 2024-09-18
Be more graceful with faulty HTML (created from famous email clients) which are interpreted by all browsers gracefully, too

carli2:
A famous email client creates quotations with `<blockquote style="width: 100%%">`. This is faulty CSS however all browsers I tested interpreted %% as % but not as px. Mpdf however defaults to "px" interpretation which leads to 100px wide blockquotes which in consequence creates PDF documents with a very small stripe of text, lots of line breaks and a large number of pages.

So we should mimic the browser's behaviour and interpret CSS measurements with %% as unit as percent rather than pixels.

### DIFF
--- a/src/SizeConverter.php
+++ b/src/SizeConverter.php
@@ -70,6 +70,7 @@ class SizeConverter implements \Psr\Log\LoggerAwareInterface
 				break;
 
 			case '%':
+			case '%%': // Issue2051
 				if ($fontsize && $usefontsize) {
 					$size *= $fontsize / 100;
 				} else {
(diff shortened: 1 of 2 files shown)


src/SizeConverter.php
```diff
@@ -70,6 +70,7 @@ public function convert($size = 5, $maxsize = 0, $fontsize = false, $usefontsize
 				break;
 
 			case '%':
+			case '%%':
```
przemo420, line 73:
```suggestion
			case '%%': // Issue2051
```

PR #2061, opened 2024-10-12
Add support for PHP 8.4.*

mkoprek:
Simply just add support for PHP 8.4.*


finwe:
How did you simply just test this, given that the stable is not even out and therefore composer cannot be run without ignores given this composer.json?

mkoprek:
Well I did not. I was sure there will be some pipeline started on this pull request for all checks.

finwe:
Meh

phpfui:
I just tested this under 8.4.0RC3 and it works fine.

Can we get a version tag for this?  Then the library will be ready to go for 8.4 release.

Thanks!

PR #2068, opened 2024-11-22
Fix nested table issues

fmjal:
- Minor version increase
- Fix issue /w nested Tables. Original issue talked about `Undefined Array key...`. This can be fixed by placing the inner table fully within <tr><td>...</td></td>, so we have cells. After that, we got `trying to access array ofset on null`. This is happening because the `$parentrow` is set to `-1`[Off by one error]. Fixed this.


finwe:
Well, were you really trying to propose this entire changelog as a PR?

PR #2074, opened 2024-12-04
Fix stroke-dasharray render

lennyrouanet:
Attribute `stroke-dasharray` can be empty.

Operands must be float-typed to avoid calculation errors.


finwe:
Please add a test case with code affected by this change. see test/Issues directory for reference.

finwe:
Thanks, quick and clean work!

PR #2077, opened 2024-12-11
Fix Undefined array key occurs in Text

benquike:
Similar issue as discussed here
https://github.com/mpdf/mpdf/discussions/1771


benquike:
can someone take a look at this?

ssieb:
I can't really look at it, but I expect you'll need a test case that demonstrates the problem.  See other PRs for how to do that.

finwe:
- It seems SetFont is not really an officially supported/documented method for mPDF, so its use can produce errors/unexpected outcomes
- The correct method for mitigating this should be ensuring the `sip` key is always defined in internal font definitions
- If the substitute font is really needed, please provide reasoning why and/or point to a precedent where the same procedure is used with other core fonts
- As @ssieb mentioned, a test case is needed representing the error this is fixing

PR #2091, opened 2025-02-25
Escape XML predefined entities

### LINKED ISSUE #2090: XMP Metadata does not escape XML predefined entities
### Guidelines

- [x] I understand that [if I fail to adhere to contribution guidelines and/or fail to provide all required details, this issue may be closed without review](https://github.com/mpdf/mpdf/blob/development/.github/CONTRIBUTING.md).

### Description of the bug

When creating a PDF/A file (and thus embedding XMP metadata), the ⁠`MetadataWriter` can generate invalid XML if the metadata contains XML predefined entities. This is because these characters need to be properly escaped. Failure to do so will result in the PDF/A file failing validation.

**Solution**: Ensure that the ⁠MetadataWriter properly escapes the XML predefined entities (`⁠&`, `⁠<`, `⁠>`, `⁠'`, and `⁠"`) with their corresponding character entity references (`⁠&amp;`, `⁠&lt;`, `⁠&gt;`, `⁠&apos;`, and `⁠&quot;`) before embedding the metadata.

### mPDF version

8.2.5

### PHP Version and environment (server type, cli provider etc., enclosing libraries and their respective versions)

8.4.4

### Reproducible PHP+CSS+HTML snippet suffering by the error

```php
// follows in a bit ...
```

coreequip:
Fixes #2090

- [x] Follows contribution guidelines
- [x] Includes a test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Bugfixes
 --------
 
 * Replace character entities with characters when processing the `code` attribute in the `<barcode />` tag
+* Escape XML predefined entities in XMP metadata (Fix for #2090)
 
 mPDF 8.1.x
 ===========================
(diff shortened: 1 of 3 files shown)


finwe:
Thanks for a superbly executed PR!

PR #2104, opened 2025-05-09
Check $GlyphClassMarks is not null before use in strpos

integer:
Without this fix is ommited:
`strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated`


finwe:
Can reopen when a reproducible test is added.

PR #2121, opened 2025-08-20
removed php 8.5 deprecation curl_close

brainfoolong:
Just fixed this deprecation

https://github.com/php/php-src/blob/5f477d9457de1d78536961fa5679b638102b6631/NEWS#L18


finwe:
And what about earlier versions of PHP? 🙃

brainfoolong:
I understand. In this case, i close this PR. I don't have time now to do this in the more proper way.

finwe:
shame, but thanks for the initial effort anyway.

ddevsr:
I create new PR #2122

PR #2124, opened 2025-08-25
refactor: PHP 8.5 Increment on non-numeric string is deprecated, use …

ddevsr:
…str_increment() instead

Reference:
- https://github.com/mpdf/mpdf/actions/runs/17197233172/job/48793490189?pr=2123

cc @finwe

### DIFF
--- a/src/Utils/Strings.php
+++ b/src/Utils/Strings.php
@@ -5,7 +5,7 @@ namespace Mpdf\Utils;
 class Strings
 {
     public static function incrementString(string $str = '')
-	{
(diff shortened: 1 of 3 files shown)


finwe:
Just use a https://packagist.org/packages/symfony/polyfill-php83 for this instead of arbitrary helper class.

finwe:
never mind, it is PHP>=7.2

finwe:
Resolved in 13ed636

ddevsr:
Better performance with `++$str` than using `str_increment`. Thats why i choose conditional PHP version

Issue #2014, opened 2024-05-24
ArrayAccess::offsetSet()" might add "void" as a native return type 


dobrjansky:
### Guidelines

- [X] I understand that [if I fail to adhere to contribution guidelines and/or fail to provide all required details, this issue may be closed without review](https://github.com/mpdf/mpdf/blob/development/.github/CONTRIBUTING.md).

### Description of the bug

Hi, in Symfony I have deprecated messages

"message": "User Deprecated: Method \"ArrayAccess::offsetSet()\" might add \"void\" as a native return type declaration in the future. Do the same in implementation \"Mpdf\\PageBox\" now to avoid errors or add an explicit @return annotation to suppress this message.",

the problem is in 
vendor/mpdf/mpdf/src/PageBox.php  \ArrayAccess
in this we have to use return type like in interface ArrayAccess

    #[\ReturnTypeWillChange]
    public function offsetSet($offset, $value)

### mPDF version

8.1

### PHP Version and environment (server type, cli provider etc., enclosing libraries and their respective versions)

8.1

### Reproducible PHP+CSS+HTML snippet suffering by the error

"message": "User Deprecated: Method \"ArrayAccess::offsetSet()\" might add \"void\" as a native return type declaration in the future. Do the same in implementation \"Mpdf\\PageBox\" now to avoid errors or add an explicit @return annotation to suppress this message.",

acasademont:
This should be fixed by https://github.com/mpdf/mpdf/pull/2132

PR #2132, opened 2025-09-10
Add return type hint annotations to avoid Symfony deprecation notices

acasademont:
We can't add the real type hints because mpdf still supports older php versions, but we can add the phpdocs with the expected return types, this should be enough to remove the deprecation warnings.

This should fix https://github.com/mpdf/mpdf/issues/2014


acasademont:
@finwe any chances of this PR getting merged? 🙏  happy to help if you need anything else

PR #2139, opened 2025-10-18
 Table Rendering: tr with page-break-{before,after} avoid will compute exact line heights to decide whether to break #2138 

carli2:
Table Rendering: tr with page-break-{before,after} avoid will compute exact line heights to decide whether to break #2138

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -22176,16 +22176,21 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 
 					// lookahead for pagebreak
 					$pagebreaklookahead = 1;
+					$pagebreaklookaheadheight = $h;
 					while (isset($table['pagebreak-before']) && isset($table['pagebreak-before'][$i + $pagebreaklookahead]) && $table['pagebreak-before'][$i + $pagebreaklookahead] == 'avoid') {
 						// pagebreak-after is mapped to pagebreak-before on i+1 in Tags/Tr.php
 						$pagebreaklookahead++;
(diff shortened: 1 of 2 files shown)


carli2:
Hi, gentle ping on this PR. I recently rebased/checked our fork against the current upstream and this change is still relevant on our side. The branch still merges cleanly, and from our testing the exact lookahead logic continues to improve table pagination for `page-break-{before,after}: avoid` rows. Could you please let me know whether you would consider reviewing/merging this, or if you would prefer a refreshed PR on top of the current `development` branch?

src/Mpdf.php
```diff
@@ -22242,16 +22242,20 @@ function _tableWrite(&$table, $split = false, $startrow = 0, $startcol = 0, $spl
...
+						$pagebreaklookaheadheight += $h2;
+
 					}
 					// corner case: if the pagelookahead is bigger than the pagesize, we break anyway, so fill up the page
-					if ($pagebreaklookahead * $maxrowheight + $extra > $pagetrigger + 0.001) {
+					if ($pagebreaklookaheadheight + $extra > $pagetrigger + 0.001) {
 						$pagebreaklookahead = 1;
+						$pagebreaklookaheadheight = $h;
 					}
-					// if we exceed page boundaries: restart table on next page before printing the line
```
finwe, line 22187: This comment is not valid, that it is removed?

carli2: Good point. I restored the explanatory comment, rebased the branch onto the current `development` branch, and force-pushed the updated single-commit branch.

PR #2154, opened 2025-12-16
Enhance external link importer

jakejackson1:
Replace rudimentary link importer with [FPDI's full-featured version](https://github.com/mpdf/mpdf/discussions/1912). Support has also been added for importing and displaying link styles. 

To prevent backwards compatibility issues, the old methods used for the link importer have been deprecated and noop'd.


JanSlabon:
LOL... I should check open PRs earlier... Say hello to my version in https://github.com/mpdf/mpdf/pull/2206.

src/Writer/MetadataWriter.php
```diff
@@ -530,41 +520,61 @@ public function writeAnnotations() // _putannots
...
+
+								foreach ($values as $name => $entry) {
+									$this->writer->write('/' . $name . ' ', false);
+									$this->mpdf->writePdfType($entry);
+								}
+
+								if (isset($pl['quadPoints'])) {
+									$s = '/QuadPoints[';
+									foreach ($pl['quadPoints'] as $value) {
+										$s .= sprintf('%.2F ', $value);
```
JanSlabon, line 568: You should sync this with the precision of the rect: `%.3F`
Otherwise it could happen that the QuadPoints are outside the `Rect` and are ignored completely.

JanSlabon (commented):
Claude is wrong with its description. The main reason why this has to be fixed, is that QuadPoints have to be within the Rect value. Due to different roundings will lead to situations where points are defined outside of the Rect.

jakejackson1:
@JanSlabon the joys of AI: tell it take make a 1 character update, and it writes a paragraph of misinformation in the commit message. Ah well. I needed to rebase the PR anyway. Should be good to go.

Issue #2145, opened 2025-11-24
Allow setting custom ImageProcessor


splitbrain:
### Please describe the new functionality as best as you can.

In the past we were able to set a custom ImageProcessorClass from the constructor (not sure if that was officially supported or if it got patched into our mpdf version a long time ago). 

New MPDF has a ContainerInterface that should make loading custom classes much easier. Unfortunately it seems not to support the ImageProcessorClass. It would be great if this could be added.

Alternatively (or in Addition): it would make things tremendously easier for downstream authors if `private` vars could instead be marked as `protected`.

See also splitbrain/dokuwiki-plugin-dw2pdf#527

PR #2164, opened 2026-01-15
Enable passing an ImageProcessor class via container

annda:
This PR solves our issue described in #2145. The AssetFetcher is not flexible enough to handle our media source, even with custom HttpClient and LocalContentLoader.

The src is a url of a fetcher script, like `https://example.dokuwiki/lib/exe/fetch.php?......` It gets interpreted as a local file '/lib/exe/fetch.php?......', which of course fopen() correctly says does not exist. In effect our custom LocalContentLoader never gets called.

### DIFF
--- a/src/ServiceFactory.php
+++ b/src/ServiceFactory.php
@@ -99,7 +99,11 @@ class ServiceFactory
 
 		$hyphenator = new Hyphenator($mpdf);
 
-		$imageProcessor = new ImageProcessor(
+		$imageProcessorClass = $this->container && $this->container->has('ImageProcessorClass')
+			? $this->container->get('ImageProcessorClass')
+			: ImageProcessor::class;
+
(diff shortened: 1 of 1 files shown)


finwe:
This is a completely inadequate change. How do you plan to ensure the passed class will be compatible with what mPDF expects? The ImageProcessor is too complex and high-level class for an interface to be extracted.

Sorry, a niche usecase such as yours does not warrant such a non-sytematic change to the codebase.

Also see the [previous reaction regarding this](https://github.com/mpdf/mpdf/issues/2145#issuecomment-3571434506).

finwe:
> Simple: we inherit from your class and overwrite just one method

How will you ensure anyone else will not pass something that is not compatible and complain about it here? This is not how software should be maintained.

I do not know exactly what you are trying to achieve and, honestly, do not have time to figure out what are you trying to do from your source code.

Issue #2165, opened 2026-01-15
Provide a PathRewriter mechanism


splitbrain:
### Please describe the new functionality as best as you can.

Instead of commenting on closed issues (#2145) and closed PRs (#2164) I am opening another issue. Previous suggestions for implementations have been rejected, so let me ask for the feature we need and let you decide how it should be implemented (we're prepared to create PR if we can).

What we need is a way to rewrite arbitrary paths for resources referenced in the source HTML **before** mpdf processes it as HTTP or LocalContent ressource.

This includes rewriting references to fetcher scripts like `https://example.dokuwiki/lib/exe/fetch.php?....` or custom protocols like `dw2pdf://mypath/to/image.jpg`. We basically want to define a function that gets the path or url to **any** resource mpdf encounters and which returns a new path or url.

We of course could preprocess the HTML, but that means the HTML is parsed twice which is a waste.

So please let us know how you would suggest to implement this.

finwe:
Seeing that `fetchRemoteContent` and `fetchLocalContent` are used only in the `AssetFetcher` itself and are public mainly for testing, it should be sufficient to extract only `fetchDataFromPath` to the interface.

finwe:
I have decided that the built-in LocalContentLoader and HttpClient are so simple to instantiate compared to the work needed to enable instantiantion of classes by class name in the container, that I'll leave the instantiation on the user side.

What remains to solve is to how to prevent clashes and weird behaviour when only the custom assetfetcher is provided in the container, and not the custom contentloader and httpclient. currently, the contentloader and httpclient are not used anywhere else than in the assetFetcher, but i'd like to have a clean path if it were necessary to change this in the future.

PR #2166, opened 2026-01-16
Custom AssetFetcher support. implements #2165

splitbrain:
This patch allows mpdf consumers to provide their own AssetFetcher implementation.


finwe (changes requested):
Other than the FQN in implements, LGTM.

PR #2173, opened 2026-02-09
fix: {PAGENO} and {nb} placeholders break text-align right/center in body content

carli2:
## Summary

- **Bug**: `{PAGENO}` and `{nb}` placeholders in body content cause `text-align: right` and `text-align: center` to miscalculate text positions. The width calculation uses the placeholder string width (8 chars for `{PAGENO}`, 4 chars for `{nb}`) instead of the actual rendered page number width, shifting text to the left by up to ~24mm.
- **Root cause**: In `finishFlowingBlock()`, `$contentWidth` and `$stringWidth` are computed via `GetStringWidth()` on text chunks that still contain placeholder strings. The alignment offset (`$empty`) is then calculated from the inflated width.
- **Fix**: Added `aliasReplaceForWidth()` that substitutes placeholders with the current page number before width calculation. The actual text in the PDF stream retains placeholders for final replacement by `aliasReplace()` during output. This mirrors how headers already handle it — they call `aliasReplace()` before `WriteHTML()`, so their widths are always correct.
- **Note**: For `{nb}` (total pages, unknown at layout time), the current page number is used as an estimate. This is exact for single-page documents and provides a much closer approximation than the 4-character placeholder width for multi-page documents.

## Test plan

- [x] New test: right-aligned `{PAGENO}/{nb}` text x-position matches equivalent static text (within 5pt)
- [x] New test: center-aligned `{PAGENO}/{nb}` text x-position matches equivalent static text (within 5pt)
- [x] New test: PDF output with placeholders is valid
- [x] New test: left-aligned text is unaffected (regression check)
- [x] Full test suite passes (694 tests, 1837 assertions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -6594,7 +6594,8 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 					$content[$k] = $chunk = str_replace(chr(173), '', $chunk);
 					$content[$k] = $chunk = str_replace(chr(160), chr(32), $chunk);
 				}
-				$contentWidth += $this->GetStringWidth($chunk, true, (isset($cOTLdata[$k]) ? $cOTLdata[$k] : false), $this->textvar) * Mpdf::SCALE;
+				$widthChunk = $this->aliasReplaceForWidth($chunk);
+				$contentWidth += $this->GetStringWidth($widthChunk, true, (isset($cOTLdata[$k]) ? $cOTLdata[$k] : false), $this->textvar) * Mpdf::SCALE;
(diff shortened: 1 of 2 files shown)


carli2:
Hi, gentle ping on this PR. I recently checked our fork against the current upstream again, and this fix is still one of the remaining deltas on our side. The branch still merges cleanly for us, and the regression tests around `{PAGENO}` / `{nb}` alignment still pass. Could you please let me know whether this approach looks acceptable, or if you would prefer changes before review?

PR #2174, opened 2026-02-10
Bug: Reset font when absolute-positioning HTML that scales the text

jakejackson1:
When using `$mpdf->WriteFixedPosHTML()` with the `auto` overflow option, the wrong font styles can be set if the text needs scaling to fit the container.

A snapshot test has been created to replicate the bug, and in the second commit the fix is included with an updated snapshot.

Here's the PDF output from the test:

- [before-snapshot.pdf](https://github.com/user-attachments/files/25196891/before-snapshot.pdf)
- [after-snapshot.pdf](https://github.com/user-attachments/files/25196892/after-snapshot.pdf)
- [snapshot-diff-overlay.png](https://github.com/user-attachments/assets/9ef9c299-8190-49a2-9aa9-9571bcf03409)

When ready to merge I can rebase the PR into a single commit. 👍


jakejackson1:
@finwe same with this one. Thanks!

PR #2176, opened 2026-02-16
fix: set PDF version 1.7 for PDF/A-2 and PDF/A-3 compliance

carli2:
## Summary
- PDF/A-2 (ISO 19005-2) and PDF/A-3 (ISO 19005-3) are based on PDF 1.7 (ISO 32000-1), but mPDF declared version 1.4. Strict validators like VeraPDF flag this.
- Adds a `/Version /1.7` entry in the document catalog for PDF/A part >= 2. This is the standard PDF mechanism for overriding the header version, and avoids timing issues with the `%PDF-` header being written before `PDFAversion` is set.
- PDF/A-1 (based on PDF 1.4) remains unaffected.

## Test plan
- [x] New test: PDF/A-1B does **not** set `/Version /1.7` in catalog
- [x] New test: PDF/A-2B sets `/Version /1.7` in catalog
- [x] New test: PDF/A-3B sets `/Version /1.7` in catalog
- [x] All existing PDFA tests still pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Writer/MetadataWriter.php
+++ b/src/Writer/MetadataWriter.php
@@ -328,6 +328,16 @@ class MetadataWriter implements \Psr\Log\LoggerAwareInterface
 	public function writeCatalog() //_putcatalog
 	{
 		$this->writer->write('/Type /Catalog');
+
+		// PDF/A-2 and PDF/A-3 are based on PDF 1.7 (ISO 32000-1).
+		// The /Version entry in the catalog overrides the header version.
+		if ($this->mpdf->PDFA && strpos($this->mpdf->PDFAversion, '-') !== false) {
+			list($part) = explode('-', $this->mpdf->PDFAversion);
+			if ((int) $part >= 2) {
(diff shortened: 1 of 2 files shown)


carli2:
Hi, gentle ping on this PR. I recently compared our fork against the current upstream again, and this PDF/A version fix is still one of the remaining deltas on our side. The branch still merges cleanly for us, and the related PDF/A tests still pass. Could you please let me know whether you would consider merging this, or if you would prefer any adjustments first?

carli2:
One more note on importance: this is relevant for e-invoicing compliance. PDF/A-3 is commonly used as the container format for electronic invoice workflows, and validators in that ecosystem tend to be strict about formal conformance details. Without the correct PDF 1.7 version signalling for PDF/A-2/3, compliant invoice documents may be flagged incorrectly even if the payload itself is otherwise fine.

PR #2178, opened 2026-02-25
Fix TypeError with `rotate: none`

derrabus:
When importing a table with `style="rotate: none;"`, mpdf currently crashes with a `TypeError`.

> Unsupported operand types: string * int

This happens in this line.

https://github.com/mpdf/mpdf/blob/b59670a09498689c33ce639bac8f5ba26721dab3/src/Mpdf.php#L26786

This PR fixes this issue for us. I've also added a test that covers my change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Bugfixes
 * Replace character entities with characters when processing the `code` attribute in the `<barcode />` tag
 * Escape XML predefined entities in XMP metadata (Fix for #2090)
 * Enable Font Subsetting by Default (Fix for #1315)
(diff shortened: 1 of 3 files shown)


derrabus:
Okay, I will adjust the PR.

While I'm at it: Supporting the angle strings `90deg` and `-90deg` would be trvial. I could also translate values like `270deg`. WDYT?

finwe:
If you are up to it, it would be great, thanks!

derrabus:
Done!

derrabus:
Done!

finwe:
I'll try this or next week.

PR #2196, opened 2026-05-12
Emit explicit /Resources dictionary for SVG Form-XObjects (PDF/A compliance)

carli2:
## Summary

PDF/A content streams that reference external objects must declare them through an explicit `/Resources` dictionary. `FormWriter::writeFormObjects()` previously omitted that dictionary for SVG Form XObjects embedded via `` tags, so validators could report missing `ExtGState` resources.

This change mirrors the existing `BackgroundWriter` resource emission and uses the `['fo']` markers set by the SVG renderer to attach the relevant `ExtGState`, `Shading`, `Font`, and `ProcSet` entries directly to the Form XObject.

## Changes

- emit an explicit `/Resources` dictionary for Form XObjects written by `FormWriter`
- include only the `ExtGState`, `Shading`, and `Font` entries that were marked as used by the SVG renderer
- add a PDF/A regression test covering an embedded SVG image in PDF/A-3 output

## Testing

- `composer test`
- `composer cs`


finwe:
Please, keep commit message in line with the rest of the history: no prefix, Capital letter beginning. 

Also, acknowledge use of AI if applicable.

PR #2203, opened 2026-06-10
Remove MBString Regex Checks

jakejackson1:
The MBString regex functions are being removed from PHP (https://wiki.php.net/rfc/eol-oniguruma). An audit of mPDF's codebase shows the last use of these functions (`mb_eregi_replace`) got replaced in mPDF v5.4, but the relevant startup checks / backup encoding process was overlooked at the time.

This PR removes these unnecessary startup checks. A separate PR for the docs will be made.


jakejackson1:
Docs update: https://github.com/mpdf/mpdf.github.io/pull/214

finwe:
Thanks!

PR #2215, opened 2026-08-21
Fix column buffer offsets when ending multi-column layout

### LINKED ISSUE #2214: Undefined array key 1 in printcolumnbuffer() when ending multi-column layout
### Guidelines

- [x] I understand that [if I fail to adhere to contribution guidelines and/or fail to provide all required details, this issue may be closed without review](https://github.com/mpdf/mpdf/blob/development/.github/CONTRIBUTING.md).

### Description of the bug

When mPDF finishes a multi-column section, `SetColumns()` calls `printcolumnbuffer()` to process the content accumulated while columns were active.

In some layouts, `printcolumnbuffer()` fails with: `Undefined array key 1`

The relevant code is:

```php
$this->ColDetails[0]['add_y'] = 0;
$last_col = 0;

// Recursively add previous column's height
for ($i = 1; $i < $this->NbCol; $i++) {
    if (isset($this->ColDetails[$i]['bottom_margin']) && $this->ColDetails[$i]['bottom_margin']) { // If any entries in the column
        $this->ColDetails[$i]['add_y'] = ($this->ColDetails[$i - 1]['bottom_margin'] - $this->y0) + $this->ColDetails[$i - 1]['add_y'];
        $last_col = $i; // Last column actually printed
    }
}
```

Later in the same method, breakpoints are processed with:

```php
$breaks[] = $rely + $this->ColDetails[$c]['add_y'] - $this->y0;
```

The problem is that a column can have registered breakpoints even when its `ColDetails[$i]['bottom_margin']` was not created. In that case, the loop above does not initialize `ColDetails[$i]['add_y']`, but the later breakpoint processing still accesses it.

In my Laravel application, the resulting warning is converted to an `ErrorException`, which interrupts PDF generation during `WriteHTML()`.

### mPDF version

v8.3.1; Also reproducible on the development branch at commit 389e19e.

### PHP Version and environment (server type, cli provider etc., enclosing libraries and their respective versions)

PHP 8.4.24 CLI on Linux (php:8.4-cli Docker image)

### Reproducible PHP+CSS+HTML snippet suffering by the error

```php
require __DIR__ . '/vendor/autoload.php';

error_reporting(E_ALL);
ini_set('display_errors', '1');

set_error_handler(function ($severity, $message, $file, $line) {
    throw new ErrorException($message, 0, $severity, $file, $line);
});

$item =
    '<h4>Voice</h4>'
    . '<p>'
    . 'Description Description Description Description '
    . 'Description Description Description Description'
    . '</p>'
    . '<br />';

$html =
    '<columns column-count="2" column-gap="15" />'
    . str_repeat($item, 7)
    . '<columns column-count="1" />';

$mpdf = new \Mpdf\Mpdf();
$mpdf->WriteHTML($html);
```

Rizzello:
Fixes #2214

`printcolumnbuffer()` could access a column offset that had not been initialized when breakpoints existed without corresponding column details.

This change initializes the cumulative column offset safely when the previous column has no `bottom_margin`.

Added regression tests for the reported case and for the related empty-column state.

Tests:
- `composer test`
- `composer cs`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Bugfixes
 * Small change to better support list-style-type on list items within tables
 * Fixed parsing sheet-size CSS property for @page
 * Conditional calls to functions removed in PHP 8.5
+* Fixed undefined array key warning when ending a multi-column layout (#2214)
 
 mPDF 8.2.x
 ===========================
--- a/src/Mpdf.php
+++ b/src/Mpdf.php
@@ -24489,9 +24489,15 @@ class Mpdf implements \Psr\Log\LoggerAwareInterface
 			$last_col = 0;
 			// Recursively add previous column's height
 			for ($i = 1; $i < $this->NbCol; $i++) {
+				// Required for columns with breakpoints but no printed entries
+				$previousBottomMargin = isset($this->ColDetails[$i - 1]['bottom_margin'])
+					? $this->ColDetails[$i - 1]['bottom_margin']
+					: $this->y0;
+
+				$this->ColDetails[$i]['add_y'] = ($previousBottomMargin - $this->y0) + $this->ColDetails[$i - 1]['add_y'];
+
 				if (isset($this->ColDetails[$i]['bottom_margin']) && $this->ColDetails[$i]['bottom_margin']) { // If any entries in the column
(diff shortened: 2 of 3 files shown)


finwe:
Thanks for the PR! I have rebased it and merged in #2220.

PR #2216, opened 2026-08-28
Add Calibri font support with regular, bold, italic, and bold italic variants

Shadow243:
Registers the Calibri TrueType font family (Regular, Bold, Italic, Bold-Italic) in mPDF's built-in font data. "calibri" was already listed in the sans_fonts substitution list, but no actual font files were provided, so any document requesting Calibri silently fell back to DejaVu Sans instead. 

This adds the missing ttfonts/calibri*.ttf files and their fontdata entry so font-family: Calibri renders with the real font.


jakejackson1:
Hey @Shadow243,

Thanks for your contribution to mPDF! Unfortunately Calibri isn't an open source font and [the license provided by Microsoft doesn't allow redistribution through a library like mPDF](https://learn.microsoft.com/en-us/typography/fonts/font-faq).

